PR #29, opened 2026-05-19
feat(ranks): wire leaderboards through maps module + LOCAL/GLOBAL tabs

GigaHierz:
## Summary

- Adapter `lib/maps/adapter.ts` converts the contract-shaped `PixelView[]` into the pure `MapSnapshot` used by the maps module (derives `(x, y)` from index, maps `ZERO_ADDRESS` to `null`, narrows the bigint price to a `number` for ranking).
- `useLeaderboard` now delegates to `allLeaderboards` / `allGlobalLeaderboards`. It exposes both `local` and `global` board sets and keeps top-level `area / empire / tycoons` aliases (LOCAL) for back-compat.
- `/ranks` page gains a `ScopeTabs` row (LOCAL / GLOBAL) above the existing metric tabs. With one revealed map both scopes show the same data; the wiring is ready to scale once `useMaps()` supplies an array of snapshots.
- New adapter tests round-trip a fake `PixelView[]` through the leaderboard module; hook tests now stub the land mask so the maps module's water filter is satisfied, and a new case verifies LOCAL == GLOBAL with one map plus on-chain profile labels applied.

## Test plan

- [x] `pnpm --filter web type-check` (clean)
- [x] `pnpm --filter web test` (23 files, 160 tests passing)
- [ ] Manual smoke on `/ranks` at 360x640: scope toggle and metric tabs both update the visible list, pixel font preserved, empty / loading states unchanged

### DIFF
--- a/apps/web/src/__tests__/hooks/useLeaderboard.test.ts
+++ b/apps/web/src/__tests__/hooks/useLeaderboard.test.ts
@@ -1,30 +1,56 @@
-import { describe, it, expect } from 'vitest'
+import { describe, it, expect, beforeAll } from 'vitest'
 import { renderHook } from '@testing-library/react'
-import { useLeaderboard } from '@/hooks/useLeaderboard'
+import { useLeaderboard, type OwnerProfileData } from '@/hooks/useLeaderboard'
 import type { PixelView } from '@/lib/mock'
-import { ZERO_ADDRESS } from '@/constants/map'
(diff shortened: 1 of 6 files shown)


GigaHierz:
Superseded by #34. Both deliver the same LOCAL/GLOBAL leaderboard wiring; #34 is the spec-aligned version dispatched after the design docs landed in #28. Closing this one to avoid two competing implementations. If we later want to extract `ScopeTabs` as its own component (the one thing #29 did that #34 didn't), that's a small refactor on top of #34.

PR #30, opened 2026-05-19
feat(storage): Postgres AssignmentStore + settings

GigaHierz:
## Summary

- Adds a Postgres-backed `AssignmentStore` + `settings` row per ADR-5: the app-layer state behind sticky wallet -> home-map placement and the operator-tunable launch knobs (`open_next_threshold`, `revealed_map_ids`).
- Uses Drizzle ORM on top of the `@neondatabase/serverless` HTTP driver. The DB is provisioned via the Vercel Marketplace (Neon) and the operator manages it via the Vercel dashboard data editor — no custom admin UI ships for launch.
- Stickiness enforced at the storage layer: `INSERT ... ON CONFLICT (address) DO NOTHING`, so an accidental second write never overwrites a home map.
- Day-one defaults baked into the settings helpers (`threshold = 2`, `revealed_map_ids = [0]`) so a fresh DB is usable without seed scripts. Schema is N-map ready.

## Files

- `apps/web/src/lib/storage/schema.ts` — Drizzle schema for `assignments` and `settings`.
- `apps/web/src/lib/storage/db.ts` — lazy Neon + Drizzle client; throws a clear error if `DATABASE_URL` is unset.
- `apps/web/src/lib/storage/assignmentStore.ts` — `PostgresAssignmentStore` (async `get`/`set`).
- `apps/web/src/lib/storage/settings.ts` — `getThreshold`/`setThreshold`/`getRevealedMapIds`/`setRevealedMapIds`.
- `apps/web/src/__tests__/lib/storage/assignmentStore.test.ts` — mock-DB tests covering sticky semantics, null lookups, case-insensitive addresses, and round-trip.
- `docs/STORAGE_SETUP.md` — Vercel Marketplace provisioning + paste-into-data-editor SQL.
- `apps/web/.env.example` — `DATABASE_URL` placeholder.
- `apps/web/package.json` — adds `@neondatabase/serverless` and `drizzle-orm`.

## Decisions

- **Drizzle vs raw pg:** chose Drizzle as the task brief requested. The schema is two tables so the cost is small and we get typed queries + clean `ON CONFLICT DO NOTHING`/`DO UPDATE` ergonomics. The README also includes raw SQL the operator pastes into the Vercel data editor, so Drizzle migrations are not on the critical path for v1.
- **`@neondatabase/serverless` (HTTP) over `pg`:** matches the Vercel Marketplace integration and works in Next.js edge/serverless runtimes without extra config.
- **Async interface widening:** `lib/maps/types.ts` `AssignmentStore` is sync. A real Postgres store cannot be sync. Rather than mutate the pure module's interface in this PR (which would touch `assignment.ts` and ripple into tests outside the brief), `PostgresAssignmentStore` exposes a separate `AsyncAssignmentStore` interface with `Promise`-returning `get`/`set`. API-route callers await it directly. A follow-up can widen the pure module if the consumer pattern justifies it.
- **Sticky `set` returns the persisted value, not the attempted one:** if a row already exists, the caller gets the stored `mapId` back. This makes the contract honest: "you asked to set X, the home map is now Y" rather than silently returning X and having a stale UI default.
- **Lazy DB client:** constructed on first call so unit tests that never touch storage need no env var; production routes pay the cost at request time, by which point env vars are populated.

## Test plan

- [x] `pnpm install` clean (added `@neondatabase/serverless`, `drizzle-orm`).
- [x] `pnpm --filter web type-check` passes.
- [x] `pnpm --filter web test` passes — 23 files / 157 tests (152 existing + 5 new).
- [ ] Manual smoke against a real Neon database (snippet in `docs/STORAGE_SETUP.md`):
  - [ ] First `set('0xtest', 0)` writes the row.
  - [ ] Second `set('0xtest', 99)` does NOT overwrite — `get` still returns `0`.
  - [ ] Delete the test row.
- [ ] Operator can edit `revealed_map_ids` JSONB in the Vercel data editor and the change is reflected on next read.

## Notes

Do not merge yet — leaving for review per the task brief. Once merged, the operator can provision Neon via the Vercel Marketplace and paste the SQL from `docs/STORAGE_SETUP.md` to create the tables.

### DIFF
--- a/apps/web/.env.example
+++ b/apps/web/.env.example
@@ -1,0 +1,13 @@
+# Mondeto — web app environment variables.
+# Copy to .env.local and fill in the values.
+
+# Public Celo RPC (mainnet)
+CELO_RPC_URL=https://forno.celo.org
+
+# RainbowKit / WalletConnect
+NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID=your_project_id_here
+
+# Postgres connection string. Provisioned via the Vercel Marketplace (Neon)
+# for production/preview; for local dev paste any Postgres URL the Neon
+# serverless driver understands. See docs/STORAGE_SETUP.md.
+DATABASE_URL=
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -13,6 +13,7 @@
   },
   "dependencies": {
     "@celo/builder-codes": "^0.2.0",
+    "@neondatabase/serverless": "^0.10.0",
     "@privy-io/react-auth": "^3.17.0",
     "@privy-io/wagmi": "^4.0.2",
     "@radix-ui/react-dialog": "^1.0.5",
@@ -21,6 +22,7 @@
     "@vercel/speed-insights": "^2.0.0",
     "class-variance-authority": "^0.7.0",
     "clsx": "^2.0.0",
+    "drizzle-orm": "^0.36.0",
     "lucide-react": "^0.460.0",
     "next": "^14.0.0",
     "obscenity": "^0.4.6",
--- a/apps/web/src/__tests__/lib/storage/assignmentStore.test.ts
+++ b/apps/web/src/__tests__/lib/storage/assignmentStore.test.ts
@@ -1,0 +1,120 @@
+/**
+ * PostgresAssignmentStore — sticky-behavior tests with a mock DB.
+ *
+ * We mock the Drizzle client surface used by the store: `.select(...)` and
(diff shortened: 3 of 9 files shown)


GigaHierz:
Superseded by #35. Both implement the Postgres-backed AssignmentStore; #35 follows the spec's file layout (`lib/maps/*` instead of `lib/storage/*`), uses raw `@vercel/postgres` instead of Drizzle (no extra ORM dep for ~30 lines of SQL), and made the `AssignmentStore` interface async — which the multi-map UI in #32 needs. Closing this one.

PR #60, opened 2026-05-22
docs(minipay): add withdraw sample tx for submission form

GigaHierz:
## Summary
Adds a withdraw sample transaction to the MiniPay Sample Transactions table. Verified on staging via #59.

## Test plan
- [x] Docs-only


GigaHierz:
Replacing with #61 (branched cleanly from main; the original branch was off staging and dragged in unrelated changes that conflicted with main).

PR #204, opened 2026-08-09
chore(minipay): diagnose why old system WebViews never draw the map (not a fix)

### LINKED ISSUE #196: MiniPay freezes and force-closes on entering Mondeto — player cannot buy at all
## Summary

2. The app freezes on entry

|---|---|---|

## Suggested investigation

## Suggested order

## Acceptance

## Status

## Note on sourcing

csacanam:
> **This does not fix #196.** It was opened as a fix, the fix turned out to be wrong, and this body has been rewritten to say what the branch actually is: a diagnosis plus two partial pieces. Draft on purpose — do not merge it expecting the map to draw.

## What this branch actually contains

1. **`lib/debugErrorOverlay.ts`** — `?debug=1` paints uncaught errors into an on-screen banner. This is what produced the answer, and it is the piece worth keeping: MiniPay's WebView is **not remotely inspectable** (with the app running, the only devtools socket on the device is Chrome's), so "see the browser console" is a dead end on exactly the devices that fail.
2. **`lib/legacyShim.ts`** + 14 tests — ES5 shim for `Object.hasOwn` and `String.prototype.replaceAll`. Correct, tested, and **currently useless**: the chunk it would help never parses.

## The real blocker, and why the shim doesn't reach it

Read off the failing device through the overlay:

```
ERROR: Uncaught SyntaxError: Unexpected token '='
  at /_next/static/chunks/286-cd92f5e4412c85c4.js:1
ERROR: Uncaught SyntaxError: Unexpected token '='
  at /_next/static/chunks/d0466ee9-966bcdea2d9ea686.js:182
```

Logical assignment operators — `||=` and `??=`, Chrome **85+**:

```js
if (e === (u ||= {})) return;   // Chrome 80 reads `u ||`, hits `=`, throws
```

| Chunk | `\|\|=` | `??=` |
|---|---|---|
| `286` (1.3 MB) | 10 | 4 |
| `d0466ee9` (405 KB) | 1 | 0 |

A syntax error kills the chunk at **parse** time, so nothing inside it runs — including the `Object.hasOwn` calls the shim was written for. I found those first with a grep for known-missing APIs, and mistook them for the cause. They are real (`Object.hasOwn` appears 10× across 4 chunks, 5 of them in chunk 286) and they will bite the moment the syntax is fixed. They are just not what breaks first.

The syntax comes from dependency `dist` code, not ours — Next down-levels our source but ships `node_modules` as published:

```
viem                    12 files
@privy-io/react-auth     3
@wagmi/connectors        1
@solana/kit / x402       transitive — source of the exact failing snippet
```

## What was tried and did not work

Recorded so the next person doesn't repeat it:

| Attempt | Result |
|---|---|
| `transpilePackages` with all five packages | 29 occurrences survive the build |
| `+ browserslist chrome >= 80` in `apps/web/package.json` | 29 survive |
| Build with `optimization.minimize = false` | 33 survive — rules out the SWC minifier introducing them |

Both config changes were reverted; they are not in this branch. Whatever fixes this needs a mechanism that actually down-levels `node_modules` — likely an explicit SWC/Babel loader over dependency code — which is a webpack change to a production app and wants a decision behind it.

## Why it's open rather than closed

The overlay is reusable and this class of bug will recur. The shim becomes necessary as soon as the syntax is resolved. The failed-attempts table above is the most expensive thing here to rediscover.

## Verification

`tsc --noEmit` clean. `pnpm --filter web test` — **417 pass** (14 new). Production build green. Preview loads with no console errors on desktop Chrome 150, and on a Pixel 9 Pro (WebView 150) — so neither piece regresses modern engines; both patches sit behind feature detection and the overlay behind `?debug=1`.

**Confirmed still broken** on the target: Huawei Mate 20 Lite (SNE-LX3), system WebView 80.0.3987.99, in MiniPay — the map still does not draw. That is the whole point of this body.

## Scope

Refs #196. The support-floor decision — how old a WebView we intend to run on — is open there and determines whether the real fix is worth building and how big it is.

### DIFF
--- a/apps/web/src/__tests__/lib/legacyShim.test.ts
+++ b/apps/web/src/__tests__/lib/legacyShim.test.ts
@@ -1,0 +1,98 @@
+import { describe, it, expect } from 'vitest'
+import vm from 'node:vm'
+import { LEGACY_SHIM } from '@/lib/legacyShim'
+
+/**
+ * The shim ships as a string inlined into <head>, so a plain import can't
+ * exercise it. Each case builds a fresh VM context, deletes the natives to
+ * simulate Chrome 80, evaluates the exact string we inline, and asserts the
+ * replacement behaves — which is also what catches a bad backslash in the
+ * escaping, the one thing a string-typed patch can silently get wrong.
+ */
+function runOnLegacyEngine(expression: string): unknown {
+  const context = vm.createContext({})
+  vm.runInContext(
+    'delete Object.hasOwn; delete String.prototype.replaceAll;',
+    context,
+  )
+  vm.runInContext(LEGACY_SHIM, context)
+  return vm.runInContext(expression, context)
+}
+
+describe('LEGACY_SHIM on an engine missing both APIs', () => {
+  it('installs Object.hasOwn', () => {
+    expect(runOnLegacyEngine('typeof Object.hasOwn')).toBe('function')
+  })
+
+  it('Object.hasOwn reports own properties', () => {
+    expect(runOnLegacyEngine('Object.hasOwn({ a: 1 }, "a")')).toBe(true)
+  })
+
+  it('Object.hasOwn ignores inherited properties', () => {
+    expect(runOnLegacyEngine('Object.hasOwn({}, "toString")')).toBe(false)
+  })
+
+  it('Object.hasOwn works on array indices', () => {
+    expect(runOnLegacyEngine('Object.hasOwn(["x"], 0)')).toBe(true)
+    expect(runOnLegacyEngine('Object.hasOwn(["x"], 1)')).toBe(false)
+  })
+
+  it('installs String.prototype.replaceAll', () => {
+    expect(runOnLegacyEngine('typeof "".replaceAll')).toBe('function')
+  })
+
+  it('replaces every occurrence, not just the first', () => {
+    expect(runOnLegacyEngine('"a-b-c".replaceAll("-", "+")')).toBe('a+b+c')
+  })
+
+  it('treats the needle literally rather than as a pattern', () => {
+    // Without escaping, "." would match every character.
+    expect(runOnLegacyEngine('"a.b.c".replaceAll(".", "-")')).toBe('a-b-c')
+  })
+
+  it('escapes regex metacharacters in the needle', () => {
+    expect(runOnLegacyEngine('"1+1=2".replaceAll("+", " plus ")')).toBe(
+      '1 plus 1=2',
+    )
+    expect(runOnLegacyEngine('"a(b)c".replaceAll("(b)", "B")')).toBe('aBc')
+  })
+
+  it('keeps $& substitution semantics', () => {
+    expect(runOnLegacyEngine('"ab".replaceAll("a", "[$&]")')).toBe('[a]b')
+  })
+
+  it('supports a function replacer', () => {
+    expect(
+      runOnLegacyEngine('"a-a".replaceAll("a", function (m) { return m.toUpperCase() })'),
+    ).toBe('A-A')
+  })
+
+  it('accepts a global RegExp', () => {
+    expect(runOnLegacyEngine('"a1b2".replaceAll(/[0-9]/g, "#")')).toBe('a#b#')
+  })
+
+  it('rejects a non-global RegExp, as the spec requires', () => {
+    // Matched by message, not `toThrow(TypeError)`: the error is constructed
+    // inside the VM realm, so it fails an instanceof against the host's.
+    expect(() => runOnLegacyEngine('"a1".replaceAll(/[0-9]/, "#")')).toThrow(
+      /must be called with a global RegExp/,
+    )
+  })
+
+  it('returns the string unchanged when the needle is absent', () => {
+    expect(runOnLegacyEngine('"abc".replaceAll("z", "!")')).toBe('abc')
+  })
+})
+
+describe('LEGACY_SHIM on a modern engine', () => {
+  it('leaves the native implementations in place', () => {
+    const context = vm.createContext({})
+    const nativeHasOwn = vm.runInContext('Object.hasOwn', context)
+    vm.runInContext(LEGACY_SHIM, context)
+    expect(vm.runInContext('Object.hasOwn', context)).toBe(nativeHasOwn)
+  })
+
+  it('is ES5-only so it cannot itself throw on the engines it targets', () => {
+    expect(LEGACY_SHIM).not.toMatch(/=>|\bconst\b|\blet\b|`/)
+  })
+})
--- a/apps/web/src/app/layout.tsx
+++ b/apps/web/src/app/layout.tsx
@@ -6,6 +6,8 @@ import { PostHogProvider } from "@/components/posthog-provider"
 import { CurrentMapProvider } from "@/hooks/useMaps"
 import { RevealsProvider } from "@/hooks/useRevealedMapIds"
 import RewardAnnouncement from "@/components/RewardAnnouncement"
+import Script from 'next/script'
+import { LEGACY_SHIM } from "@/lib/legacyShim"
 
 const APP_URL = 'https://www.mondeto.app'
 const TITLE = 'Mondeto — every pixel is up for grabs'
@@ -66,6 +68,18 @@ export default function RootLayout({
   return (
     <html lang="en">
       <head>
+        {/* Built-ins that our dependency tree calls but pre-2021 Chromium
+            lacks. `beforeInteractive` is what puts this ahead of Next's own
+            bootstrap scripts — a plain inline <script> here is emitted
+            *after* them in the head, and we need the patches installed
+            before any chunk evaluates. MiniPay renders with the device's
+            system WebView, which on older Android is still Chrome 80. See
+            lib/legacyShim.ts. */}
+        <Script
+          id="legacy-shim"
+          strategy="beforeInteractive"
+          dangerouslySetInnerHTML={{ __html: LEGACY_SHIM }}
+        />
         {/* Pre-warm Google Fonts DNS + TLS so the @font-face requests don't
             block first paint. Combined with preload below this is the
             highest-impact PageSpeed change for our mobile target. */}
--- a/apps/web/src/app/layout.tsx
+++ b/apps/web/src/app/layout.tsx
@@ -8,6 +8,7 @@ import { RevealsProvider } from "@/hooks/useRevealedMapIds"
 import RewardAnnouncement from "@/components/RewardAnnouncement"
 import Script from 'next/script'
 import { LEGACY_SHIM } from "@/lib/legacyShim"
+import { DEBUG_ERROR_OVERLAY } from "@/lib/debugErrorOverlay"
 
 const APP_URL = 'https://www.mondeto.app'
 const TITLE = 'Mondeto — every pixel is up for grabs'
@@ -79,6 +80,15 @@ export default function RootLayout({
           id="legacy-shim"
           strategy="beforeInteractive"
           dangerouslySetInnerHTML={{ __html: LEGACY_SHIM }}
+        />
+        {/* `?debug=1` only — paints uncaught errors into an on-screen banner.
+            MiniPay's WebView has no reachable console, so "see the browser
+            console" is a dead end on the devices that actually fail. See
+            lib/debugErrorOverlay.ts. */}
+        <Script
+          id="debug-error-overlay"
+          strategy="beforeInteractive"
+          dangerouslySetInnerHTML={{ __html: DEBUG_ERROR_OVERLAY }}
         />
         {/* Pre-warm Google Fonts DNS + TLS so the @font-face requests don't
             block first paint. Combined with preload below this is the
--- a/apps/web/src/lib/debugErrorOverlay.ts
+++ b/apps/web/src/lib/debugErrorOverlay.ts
@@ -1,0 +1,51 @@
+/**
+ * On-screen error reporter for environments with no reachable console.
+ *
+ * MiniPay renders miniapps in the device's Android System WebView, and that
+ * WebView is not remotely inspectable — with MiniPay running, the only
+ * devtools socket on the device is Chrome's. So when the app dies there,
+ * Next's "a client-side exception has occurred (see the browser console)"
+ * is a dead end: there is no console to see.
+ *
+ * This installs `onerror` + `unhandledrejection` handlers that paint the
+ * message, source and line into a fixed banner, so the failure is readable
+ * on the device itself.
+ *
+ * Opt-in via `?debug=1` — it never runs for normal traffic. Inlined right
+ * after the legacy shim (see `app/layout.tsx`) so it is armed before any
+ * chunk evaluates and catches failures during bundle execution, which is
+ * exactly where old engines break.
+ *
+ * ES5 only, same constraint as `legacyShim.ts`: this has to run on the
+ * engines it is meant to diagnose.
+ */
+export const DEBUG_ERROR_OVERLAY = [
+  '(function(){',
+  'if(String(location.search).indexOf("debug=1")===-1){return}',
+  'var box=null;',
+  'function show(kind,msg,src,line){',
+  'try{',
+  'if(!box){',
+  'box=document.createElement("div");',
+  'box.setAttribute("style","position:fixed;left:0;right:0;top:0;z-index:2147483647;'
+    + 'max-height:60%;overflow:auto;background:#1B1B1B;color:#A7FF05;'
+    + 'font:11px/1.45 monospace;padding:8px;border-bottom:2px solid #FF4C00;'
+    + 'white-space:pre-wrap;word-break:break-word");',
+  '(document.body||document.documentElement).appendChild(box)}',
+  'var e=document.createElement("div");',
+  'e.setAttribute("style","margin-bottom:8px");',
+  'e.appendChild(document.createTextNode(',
+  'kind+": "+msg+(src?("\\n  at "+src+(line?(":"+line):"")):"")));',
+  'box.appendChild(e)',
+  '}catch(ignored){}}',
+  'window.onerror=function(msg,src,line){show("ERROR",msg,src,line)};',
+  'window.addEventListener("unhandledrejection",function(ev){',
+  'var r=ev&&ev.reason;',
+  'show("UNHANDLED REJECTION",(r&&(r.stack||r.message))||String(r),"",0)});',
+  // Engine report, so a failure comes with the environment that produced it.
+  'window.addEventListener("DOMContentLoaded",function(){',
+  'show("UA",navigator.userAgent,"",0);',
+  'show("HAS",("hasOwn:"+(typeof Object.hasOwn)+" replaceAll:"',
+  '+(typeof String.prototype.replaceAll)),"",0)});',
+  '})();',
+].join('')
--- a/apps/web/src/lib/legacyShim.ts
+++ b/apps/web/src/lib/legacyShim.ts
@@ -1,0 +1,49 @@
+/**
+ * Inline shim for JS built-ins that our dependency tree calls but that
+ * pre-2021 Chromium does not have.
+ *
+ * Why this exists: MiniPay renders every miniapp with the *device's*
+ * Android System WebView, which is a system component the user neither
+ * updates nor knows about. On a Huawei Mate 20 Lite with Play Store and
+ * full Google services installed, that engine is still the factory
+ * 80.0.3987.99 (Chrome 80, Feb 2020) while the Chrome app on the same
+ * phone is 150 — so Mondeto loads in Chrome and shows a bare blue ocean
+ * in MiniPay. `Object.hasOwn` (Chrome 93+) throws there, which is enough
+ * to stop `fetchAllPixelsFromContract` from ever resolving, so
+ * `usePixelMap` never leaves 'loading' and `PixelLayer` never draws.
+ *
+ * This runs as a blocking inline script in <head> (see `app/layout.tsx`)
+ * so it is installed before any Next chunk evaluates. It must therefore
+ * stay ES5 — no arrow functions, no const/let, no template literals —
+ * and small enough that inlining it costs less than a round trip.
+ *
+ * Each patch is feature-detected, so modern engines pay nothing but the
+ * two `if` checks.
+ *
+ * The source lives here as a string rather than a module because an
+ * inline <script> cannot import. `__tests__/lib/legacyShim.test.ts`
+ * evaluates this exact string with the natives deleted and asserts the
+ * replacements behave, which is what keeps the escaping honest.
+ */
+export const LEGACY_SHIM = [
+  '(function(){',
+  // Object.hasOwn — Chrome 93. Called ~10x across our chunks, including
+  // the deep-equality walk that runs on every wagmi/react-query read.
+  'if(!Object.hasOwn){Object.defineProperty(Object,"hasOwn",{',
+  'value:function(o,k){return Object.prototype.hasOwnProperty.call(Object(o),k)},',
+  'configurable:true,writable:true})}',
+  // String.prototype.replaceAll — Chrome 85.
+  'if(!String.prototype.replaceAll){Object.defineProperty(String.prototype,"replaceAll",{',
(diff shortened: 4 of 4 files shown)


GigaHierz:
Agreed this should stay a draft, and agreed the overlay is the piece worth keeping — the failed-attempts table alone justifies the branch existing. Two notes, one of which changes the overlay before it's worth extracting.

### The `beforeInteractive` ordering claim is inverted

`layout.tsx` says:

> `beforeInteractive` is what puts this ahead of Next's own bootstrap scripts — a plain inline `<script>` here is emitted *after* them in the head, and we need the patches installed before any chunk evaluates.

In App Router it does the opposite. From `next@15.5.20`:

- `next/dist/client/script.js` (~L274–290) — an **inline** `beforeInteractive` script is not emitted as an executable inline script at all. It emits a queue push:
  ```js
  (self.__next_s = self.__next_s || []).push([0, { children: "<our source>", id: "legacy-shim" }])
  ```
  That only stores the source **as a string**. Nothing executes.
- `next/dist/client/app-bootstrap.js` (~L22–53) — `appBootstrap()` calls `loadScriptsInSequence(self.__next_s, hydrate)`, which walks the queue, does `el.innerHTML = props.children`, and appends to `<head>`.

So the real order is:

```
HTML parse → queues a string (nothing runs)
  → main-app.js downloads, parses, executes
    → appBootstrap drains __next_s → shim + overlay finally run
      → hydrate
        → page chunks (286, d0466ee9, …)
```

The patches install **after** Next's runtime, not before it — and `appBootstrap` is itself inside a chunk that has to parse first.

The consequence that matters is for the overlay, not the shim: **as written it cannot catch a parse error in `main-app.js`, the framework chunk, or the polyfills chunk.** It caught the errors in this investigation because `286` and `d0466ee9` load after hydration begins — that's chunk-placement luck, not design. The failure modes where the overlay is *most* needed are exactly the ones it currently can't see.

A plain inline `<script dangerouslySetInnerHTML>` in `<head>` is strictly better for both purposes: it executes synchronously during HTML parse, ahead of every Next script (all of which are `async` / `defer` / module-preloaded). No queue, no dependency on the runtime booting.

Worth verifying either way by viewing source on the preview and checking whether `<head>` contains the shim body or a `self.__next_s` push — that's a five-second check and it settles it.

### Suggested split

- **Extract the overlay into its own small PR** with the plain inline `<script>`, and merge that. It's independently useful, this class of bug will recur, and it's the only tool we have on a surface with no reachable console. It shouldn't be gated behind an unsolved fix.
- **Leave the shim here.** Your read is right that it's correct-but-unreachable today; it becomes load-bearing the moment the syntax problem is solved, and the tests are worth keeping alive on the branch until then.

### Context that affects the next step

The support floor is now decided — **Chrome 80** — so the dependency down-levelling work is in scope. Details in #196, plus a correction to the source list there: `@celo/attribution-tags` isn't a source (its only deps are `ox` and `viem`), and the Solana/x402 subtree arrives as a non-optional dependency of `@privy-io/react-auth`, which is lazy-loaded and never runs for MiniPay users.

That last point is worth a look before you scope the transpile pass: the `(u ||= {})` snippet you quoted traces to `@solana/offchain-messages` / `@solana/transactions`, which per `wallet-provider.tsx` shouldn't be reachable on a MiniPay load at all. If it's genuinely in the failing chunk, the lazy-Privy split is leaking — which would be its own bug, and a meaningful bundle-size win on these devices independent of the parse issue. `viem` (12 files) is on the MiniPay path regardless and is the one the fix has to cover.

GigaHierz:
Second review pass. Agreed this stays a draft, and agreed the overlay is the piece
worth keeping — so I've aimed at the questions the thread leaves open rather than
re-deriving the `beforeInteractive` / `__next_s` ordering, which I take as read and
extend below.

Two direct answers up front, both clean: **no XSS**, and the tests **do** genuinely
exercise the shim rather than silently testing Node's natives. Details in Strengths.

### Strengths

- **The overlay is written correctly at the DOM level.** `debugErrorOverlay.ts:35-39`
  builds each entry with `document.createElement` + `document.createTextNode` and a
  *static* `setAttribute("style", …)`. There is **no `innerHTML`, no
  `insertAdjacentHTML`, no `document.write`** anywhere in it. The error message, the
  `src`, the `line`, the rejection `stack` and `String(r)` all land in a text node, so
  attacker-influenced content in an error message (a URL, a thrown object with a
  hostile `toString`) cannot become markup. Worth saying explicitly, because moving to
  a plain inline `<script dangerouslySetInnerHTML>` makes this property load-bearing.
- **The tests genuinely exercise the shim.** This is the classic polyfill-test failure
  mode and this PR does not fall into it. `legacyShim.test.ts:12-19` builds a fresh
  `vm.createContext({})` — a separate realm with its own `Object` and
  `String.prototype` — deletes the natives *in that realm*, then evaluates
  `LEGACY_SHIM` there, so the `if(!Object.hasOwn)` guard takes the install branch and
  every assertion runs against the shim. `runOnLegacyEngine` also correctly matches
  the TypeError by message rather than `instanceof` (`:76-78`), because the error is
  constructed in the guest realm. A materially better harness than most hand-rolled
  polyfills ship with.
- **`replaceAll` is more spec-correct than hand-rolled shims usually are.** I worked
  through the cases you'd expect to be wrong and they're right: a non-global RegExp
  throws TypeError (`:40`); a global RegExp delegates to `replace` and inherits
  `lastIndex` reset; a function replacer receives `(match, offset, string)` natively;
  `` $&, $`, $', $$ `` all work because it delegates rather than taking the
  `split/join` shortcut; `$1` stays literal for a string needle because the escaped
  pattern has no capture groups; and the empty-string case is correct —
  `new RegExp("","g")` yields `/(?:)/g`, so `'ab'.replaceAll('','-')` returns `'-a-b-'`
  with no infinite loop. The metacharacter escape at `:45` is correct through both
  levels of escaping.
- **`Object.defineProperty`, not plain assignment** (`:32-34`, `:36-47`), with
  `configurable:true, writable:true` and `enumerable` defaulting to false — matching
  native descriptors. A plain `String.prototype.replaceAll = fn` would have made the
  method enumerable and broken every `for…in` over a string-keyed object.
- **The failed-attempts table is the most valuable artefact here** and the instinct to
  record it is right. It's also, as of today, wrong in a way that matters — C2.

### Issues

#### Critical

**C1. The diagnosis is measured against a tree that no longer exists. The branch is 8
commits behind `main` and predates #209, which removed the exact code it blames.**

`git merge-base` puts this branch on `4e8fcde` (#186); `main` is now `fe764bd`. In
between sits `bbc1801` — *"perf(minipay): stop shipping Privy, x402 and @solana/kit to
clients that never use them (#209)"*. That commit's message states, measured: logical
assignment operators on the route **fall from 21 to 6**, *"No chunk on any route now
carries Privy, x402 or Solana code"*, and *"the snippet reported as throwing on a
WebView 80 device traces to `@solana/offchain-messages`, which is now off the path
entirely."*

Every quantitative claim in the body is therefore stale:

| PR body says | On `main` today |
|---|---|
| chunk `286`: 10 `\|\|=` + 4 `??=`; `d0466ee9`: 1 | those chunk hashes no longer exist; route total is 6 |
| `@solana/kit` / `x402` — *"source of the exact failing snippet"* | not in any chunk on any route |
| `@privy-io/react-auth` 3 files | not in the MiniPay-path chunk |
| `transpilePackages` with all five → 29 survive | two of the five no longer ship at all |

The `viem` row (12 files) survives, and per #209 so do `@noble/hashes` and
`@wagmi/connectors`. **Those three are the entire remaining problem, and all three are
genuinely required** — a much smaller and more tractable target than the body
describes.

Rebase and re-measure before this body is used as a record. The "confirmed still
broken on the target device" line also needs re-testing: if it ran against this
branch's preview, it tested a pre-#209 bundle and says nothing about `main`. (#209
predicts it *is* still broken, but for a smaller reason.)

**C2. The `transpilePackages` row is under-scoped by construction, not a result. As
written it will steer the next person away from the approach most likely to work.**

`transpilePackages` matches the **named package only**, not that package's own
dependencies — it builds a path matcher against `node_modules/<name>/`, and a sibling
under a different name is untouched. Verified from `pnpm-lock.yaml`:

- `@solana/kit@5.5.1` is a **barrel** over 22 separately-published sibling packages
  (lockfile L4097-4118), one of which is `@solana/offchain-messages` — the package #209
  identifies as the source of the exact `(u ||= {})` snippet. `transpilePackages:
  ['@solana/kit']` covers **1 of 45** `@solana/*` packages in the tree and specifically
  does **not** cover `@solana/offchain-messages`.
- `viem@2.23.2`'s dependencies are `@noble/curves`, `@noble/hashes`, `@scure/bip32`,
  `@scure/bip39`, `abitype`, `isows`, `ox`, `ws` (L13421-13430). `transpilePackages:
  ['viem']` covers none of them — including `@noble/hashes`, which #209 names as still
  shipping above-80 syntax.

So "29 occurrences survive" is **exactly what a correctly-functioning
`transpilePackages` produces** when the packages that actually publish the syntax
aren't named. The attempt didn't fail; it was aimed at the wrong five packages.

The browserslist row is misleading for a different reason: Next's default client target
is already well below Chrome 80, so adding `chrome >= 80` *raises* the floor rather
than lowering it — and browserslist has no effect on `node_modules` at all unless those
modules are routed through SWC, which only `transpilePackages` arranges. That row
couldn't have changed anything by construction.

Rewrite both rows as *"attempted, misapplied"* rather than *"tried, does not work"*.
The honest open question is whether a **blanket** down-level of all dependency code to
the browserslist target works — a different mechanism (a webpack rule over
`node_modules`), and the thing that hasn't actually been tried.

**C3. The disproven causal chain is still asserted as fact inside the committed code.**
The body retracts the `Object.hasOwn` theory; `legacyShim.ts:11-13` does not:

> `* in MiniPay. Object.hasOwn (Chrome 93+) throws there, which is enough`
> `* to stop fetchAllPixelsFromContract from ever resolving, so`
> `* usePixelMap never leaves 'loading' and PixelLayer never draws.`

That's the wrong diagnosis, stated in the file's own "Why this exists". PR bodies are
read once; docblocks are read forever. The same applies to the ordering claim — it's
also in `legacyShim.ts:15-16` and `debugErrorOverlay.ts:14-17`, so that fix touches
three places, not one.

#### Important

**I1. The lazy-Privy leak was real — closing the open question above.** At this
branch's base (`4e8fcde`): `connect-button.tsx` line 3 imports `useConnectWallet` from
`@privy-io/react-auth`, and line 10 imports `PrivyReadyContext` from
`wallet-provider-privy`, which itself imports Privy. `TopBar.tsx:5,62` and
`navbar.tsx:14` mount `ConnectButton` on **every page including the map**. Webpack
resolves static imports at build time, so the runtime `privyReady` guard could not
prevent the graph landing in the shared chunk. `@privy-io/react-auth@3.17.0` carries
`x402` and `@solana/kit` as **non-optional `dependencies`** (lockfile L2280, L2304 —
`@solana/kit` is *also* an optional peer, which is a red herring: it's in
`dependencies` regardless).

Not `wallet-provider.tsx`'s fault — that file splits Privy correctly and always did.
Already fixed on `main` by #209. Also confirming the other correction:
`@celo/attribution-tags@0.3.0`'s only dependencies are `ox` and `viem` (lockfile
L333-335), so it is not a source.

**I2. `window.onerror` is the wrong hook for the failure mode this exists to catch, and
it can be clobbered.** `debugErrorOverlay.ts:41` assigns `window.onerror = …`:

- **A chunk that fails to *download* produces no overlay output at all.** Resource-load
  errors (a 404 or TLS/network failure on `/_next/static/chunks/*.js`) fire an `error`
  event on the element that does not bubble; `window.onerror` never sees it. Blank map,
  empty overlay — the exact silent failure the tool was built to break, and on a 2018
  device on a bad mobile network at least as likely as a parse error.
  `window.addEventListener('error', handler, true)` catches both, because capture-phase
  listeners on `window` see non-bubbling resource errors.
- **Assignment doesn't compose.** Anything that later sets `window.onerror` silently
  replaces the handler; anything set before is silently discarded.

**I3. The `DOMContentLoaded` engine report is a race that will frequently never fire.**
`debugErrorOverlay.ts:46-49` registers the UA / feature-detection report on
`DOMContentLoaded`. Given the `__next_s` drain ordering established above, the listener
is frequently attached **after** `DOMContentLoaded` has fired — so the UA and
`hasOwn`/`replaceAll` lines never print, non-deterministically. That's the half of the
output telling you *which engine* produced the error, i.e. the half you can't
reconstruct afterwards. Fix:
`if(document.readyState!=="loading"){report()}else{window.addEventListener("DOMContentLoaded",report)}`.

**I4. Unbounded DOM growth — the diagnostic degrades the device it's diagnosing.**
`debugErrorOverlay.ts:39` appends a new `<div>` per error with no cap and no dedupe.
The app polls on an interval; one repeatedly-rejecting promise appends an entry every
tick, forever. On a Mate 20 Lite that becomes an unresponsive page and plausibly an OOM
in the WebView — losing the session you were trying to read. Cap at ~50, or dedupe
identical `kind+msg` with a `×N` counter.

**I5. `Object.hasOwn(null, k)` returns `false` where the spec throws `TypeError`.**
`legacyShim.ts:33` — `Object.prototype.hasOwnProperty.call(Object(o), k)`. `Object(null)`
returns `{}`; it doesn't throw. Per spec `Object.hasOwn` performs `ToObject(O)`, which
throws for `null`/`undefined`. Everything else is correct — primitives coerced, symbol
keys, numeric-string keys, prototype chain correctly not consulted, `length` is 2.

This is the only real spec deviation in either polyfill. It doesn't bite today because
the shim never runs — but when it does, a dependency passing an accidental `null` gets
a quiet wrong `false` instead of a loud TypeError, on the one device class where you
can't attach a debugger. One line:
`if(o==null){throw new TypeError("Cannot convert undefined or null to object")}`.

**I6. Two shimmed APIs won't carry a Chrome 80 floor.** `Object.hasOwn` and
`String.prototype.replaceAll` were chosen by grepping one build. The Chrome 80 runtime
gap is much wider, and these are common in the viem/wagmi/noble surface #209 says is
genuinely required: `Array.prototype.at` (92), `Promise.any` (85), `Error` `cause` (93),
`structuredClone` (98), `Array.prototype.findLast` (97), `crypto.randomUUID` (92),
`AbortSignal.timeout` (103), `Object.hasOwn` (93). The floor is razor-thin in the other
direction too — `?.` and `??` are Chrome **80 exactly**, so there's zero headroom above
the target. Now the floor is settled at 80, the durable mechanism is `core-js@3` driven
by an explicit browserslist, not a hand-maintained two-function string — otherwise the
next Renovate bump silently reintroduces the problem.

#### Minor

**M1. `debugErrorOverlay.ts` has zero tests, while the piece called unreachable has
fourteen.** The overlay is the artefact recommended for merge and it's a
hand-concatenated JS string inlined into HTML — exactly the category the shim's suite
exists to protect (`legacyShim.ts:24-26` says so). It carries its own escaping hazard
at `:38` (which I checked and is correct) with nothing asserting it stays correct. At
minimum mirror the two meta-tests: an ES5 assertion, and a `vm` run firing a synthetic
error and checking the banner text.

**M2. The ES5 assertion is a weak proxy.** `legacyShim.test.ts:95` catches three
constructs and misses spread, default parameters, shorthand methods, `for…of`, class
syntax, `??`, `?.` and `||=`. Since the whole PR is about a parse error, this test
should parse rather than pattern-match. Both strings do currently pass a manual ES5
read.

**M3. `?debug=1` gating sits outside the repo's convention and the PR doesn't argue the
exception.** `apps/web/CLAUDE.md` gates dev surfaces via `VERCEL_ENV === 'production'`
→ `notFound()`. This ships to every production visitor with a user-guessable trigger. I
do **not** think a hard production gate is right — MiniPay users hit prod, so gating it
out defeats the tool — but the exception should be *stated* next to the `/dev` rule so
the next reader doesn't read it as a violation, and it's worth driving from a
`NEXT_PUBLIC_` flag so it can be switched off without a deploy (consistent with
`NEXT_PUBLIC_REVEALED_MAP_IDS` as a rollout knob). The substring match at `:24` also
matches `?nodebug=1` and `?xdebug=13`; `URLSearchParams` isn't available at ES5, so a
`/[?&]debug=1(&|$)/` test is the cheap tightening.

**M4. How does a field tester get `?debug=1` into a MiniPay session?** MiniPay launches
mini-apps from a fixed registered URL. If a tester can't append a query string, the
overlay is only reachable from the system browser — which isn't where the bug
reproduces. Worth stating how the URL was actually loaded on the Mate 20 Lite, because
that's the reproduction recipe the next person needs.

**M5. This branch predates the coverage gate (#211), so its green CI isn't meaningful
yet.** `vitest.config.ts` on `main` now enforces `src/lib/**` at 71% statements with
`include: ['src/**']`, counting untested files as zero. `legacyShim.ts` is covered;
`debugErrorOverlay.ts` is imported by no test and will count 0% against that floor.
Small impact, but the "417 pass, build green" line was measured before the gate existed.

**M6. No `</script>` guard on the inlined strings.** Neither constant contains `</`
today, so nothing is broken — but both are interpolated raw into HTML via
`dangerouslySetInnerHTML`, and a future edit adding a string containing `</script>`
would terminate the tag. A one-line `.replace(/<\//g,'<\\/')` at the injection site
makes that unfalsifiable.

### Recommendations

1. **Rebase onto `main` first, then re-measure.** Nothing else can be evaluated until
   then — the commit this branch is behind by is the one that deleted the code it
   blames. Re-run the overlay on the Mate 20 Lite against a post-#209 preview and record
   the new chunk names and counts. Expectation, from #209's own numbers: still broken,
   now attributable to `viem` / `@noble/hashes` / `@wagmi/connectors`, ~6 occurrences
   rather than 15.
2. **Split the overlay into its own small PR and merge it** — with the plain inline
   `<script dangerouslySetInnerHTML>` in `<head>`, plus I2, I3, I4 and M1. Scope it as
   `feat(minipay): on-screen error overlay for WebViews with no reachable console` and
   keep the diagnosis out of it.
3. **Keep the shim here, but fix I5 now** while the reasoning is fresh, and rewrite the
   docblock (C3) so it doesn't carry the retracted theory. Don't merge it: two functions
   won't cover a Chrome 80 floor (I6), and merging creates the impression the runtime gap
   is handled.
4. **Correct the failed-attempts table (C2) before anyone acts on it.** Single
   highest-value edit available here.
5. **The real work is now a blanket down-level, and it needs a gate to stay fixed.** A
   webpack rule running SWC over `node_modules` with an explicit `chrome 80` target, plus
   `core-js` for the runtime half. And answering "one pass or whack-a-mole" directly:
   **add a build-time assertion that greps `.next/static/chunks/*.js` for post-ES2020
   syntax and fails the build**, analogous to the source-level import guard #209 added.
   Without it, any Renovate bump reintroduces the problem silently. That check is the
   difference between a fix and a permanent maintenance burden.
6. **Post the corrected findings into #196** regardless of what happens to the branch.
   The provenance chain, the leak-and-fix history, and the `transpilePackages` scoping
   trap are the expensive things to rediscover.

### Assessment

**Ready to merge?** No — and correctly so, it's a draft. But the right action isn't
"leave it open": **rebase, correct the record, split the overlay out to merge on its
own, and park the shim.**

**Reasoning:** The method was sound and the overlay is a genuinely good instrument —
but the branch is 8 commits behind `main` and predates #209, which deleted the
dependency subtree this PR names as the cause, so the body's counts, chunk hashes and
dependency table describe a bundle that no longer ships. The one artefact meant to
outlive the branch — the failed-attempts table — is the part most in need of
correction.

---

*Verification gap worth stating: there's no build artefact in the workspace I reviewed
from, so I could not grep actual chunk output for other post-Chrome-80 syntax. The I6
list is derived from the packages' published targets plus known Chrome versions, not
from the bytes. One thing I can state structurally: **top-level await is a non-risk**
regardless, because webpack transforms TLA into async module wrappers. Static blocks
(94) and the RegExp `d` flag (90) are the real second-wave candidates and I could not
rule them in or out — someone with `node_modules` installed should grep
`node_modules/{viem,@noble,@wagmi,ox,abitype}/**/dist/**/*.js` for `static\s*{`, `d`-flag
regexes, and `\?\?=|\|\|=|&&=`. That grep is the missing evidence for the
one-pass-or-whack-a-mole question.*

GigaHierz:
Splitting this branch's two halves so neither is lost while the PR stays draft:

- **Error overlay** (`lib/debugErrorOverlay.ts` + the `beforeInteractive` wiring in `layout.tsx`) → **#247**, pinned to this PR's head SHA `b243bedb21872a0e962839cd95b86ffa51c46c50`. It is the piece that produced the answer and it is reusable.
- **`legacyShim.ts`** → superseded by **#225**, which argues a hand-written two-function shim is the wrong shape and `core-js@3` driven by browserslist is the durable one.

Also carried into #225, because it is the most expensive thing here to rediscover: the **`transpilePackages` row of the failed-attempts table above is misleading**. The mechanism works — it was aimed at package names that do not publish the failing code. `transpilePackages` matches the named package only, not its dependencies, so naming `@solana/kit` (a barrel over 22 separately-published siblings) misses `@solana/offchain-messages`, and naming `viem` misses `@noble/hashes`. "29 survive" is exactly what a correctly-functioning `transpilePackages` produces when the packages that actually publish the syntax are not named. Treat that row and the `browserslist` row as *attempted, misapplied* — not as ruled out.

No change requested to this PR; it stays draft as intended.

csacanam:
Closing this. It did its job, and that job was never merging.

This branch was opened as a fix for #196, the fix turned out to be wrong, and I rewrote the body to say so rather than quietly reworking it. What it became instead was the diagnosis — and the diagnosis is what unblocked the two PRs that actually fix the thing.

WHY IT IS NOT MERGING

The shim half is superseded. It patched Object.hasOwn and String.prototype.replaceAll, which are real gaps, but they sit behind a parse error — the chunk never runs, so the shim never gets the chance to help. #253 covers both APIs inside a core-js set derived from browserslist, so a hand-maintained 49-line shim is now the worse version of a solved problem.

The overlay half is already carried forward, on a clean branch off main, without the shim.

The branch is being kept, not deleted. Everything below is pinned to head b243bedb21872a0e962839cd95b86ffa51c46c50.

WHERE THE VALUE WENT

1. lib/debugErrorOverlay.ts -> #252

   Lifted out onto a branch off main. That PR pins this commit in its body, so the provenance is checkable after this branch is gone. It also improved on what is here in two ways worth knowing: the gate is now an exact-parameter match (/[?&]debug=1(&|$)/) rather than an indexOf, which closed ?debug=10 and ?xdebug=1; and it mounts as a raw <script> rather than next/script with beforeInteractive.

   That second change contradicts what this PR did, and it is correct. Measured in #252: in the App Router, beforeInteractive emits no script — it serialises into a self.__next_s queue that main-app drains right before hydration, so the handlers install only after the webpack runtime, React and main-app have all parsed. Poisoning any of those three left the banner empty. The version on this branch worked on the Mate 20 Lite only because the chunks that failed there (286, d0466ee9) are page-level and arrive after bootstrap — it would have been blind to a failure in the bootstrap path itself.

   So: the overlay survives, and the one thing this branch got wrong about it has been found and fixed with evidence.

2. lib/legacyShim.ts + its 14 tests -> superseded by #253 / #225

   Not carried forward, deliberately. See above. Recoverable from this branch if core-js ever turns out to miss something, but it should not be resurrected by default.

3. The failed-attempts table and the device evidence -> captured on #225

   This is the part that would actually have been expensive to lose, so it is written out in full on the issue rather than left in a closed PR body: the three mechanisms that do not work with their measured counts, why each fails, the two failing chunks with their operator counts, and the reproducing handset.

   It was not just archived — #253 used it. This branch measured that transpilePackages leaves 29 occurrences alive, and #253 opens by making exactly that argument for a blanket SWC pass over node_modules instead of the named-package option. The negative result shaped the fix.

THE DEVICE, BECAUSE TWO OPEN BOXES STILL NEED IT

Huawei Mate 20 Lite (SNE-LX3), Android System WebView 80.0.3987.99, in MiniPay. That is the handset that reproduces #196 and it is at the support floor, not below it.

Both #253 (#225 box: verified on a real device at the floor) and #252 (#247 box 4: load ?debug=1 on a floor device) are gated on a check against a WebView at or below 84. It is one session on one handset, and #252's overlay is the tool for reading the result. Whoever picks that up: this is the device.

Refs #196, #225, #247. Superseded in part by #252 and #253.

PR #205, opened 2026-08-09
feat(web): count the browser engines that never reach analytics

csacanam:
## Problem

#196 opens by pointing out that we can't see the bug it reports:

> It is also invisible to our analytics: a client that hard-freezes on entry may never fire a single PostHog event, so the affected population could be much larger than two.

Investigating that issue turned the worry into a certainty. On a Huawei Mate 20 Lite whose system WebView is still the 2018 factory build (Chrome 80), our chunks fail to **parse** — dependencies publish `||=` and `??=`, which are Chrome 85+. A parse error kills the chunk before a single line runs, so PostHog never initialises. It isn't that these users produce fewer events; they produce none. Every browser we most need to count is structurally absent from client analytics.

So today the honest answer to "is this two users or two thousand?" is that we have no way to know.

## What changes

The document request happens **before** any of our JavaScript, so read it there.

- **`lib/userAgentInsight.ts`** — pure parser returning the Chromium major, whether the UA is an embedded Android WebView (the `; wv)` token), and whether it's below the version our current bundle needs. Dependency-free so it can be tested against real UA strings.
- **`app/layout.tsx`** — reads the `user-agent` header and emits one `logger.info('document request', …)` per render with those fields alongside the raw string.

Logging the parsed fields rather than only the raw UA is what makes this queryable: the distribution comes out of a group-by instead of eyeballing strings.

`KNOWN_PARSEABLE_CHROME_MAJOR` is set to 85 because that is the syntax our dependencies ship, **not** because we've decided to support Chrome 85. The support-floor decision is open in #196; this constant should follow it once it's settled, not pre-empt it.

## Why it's safe

`layout.tsx` already carries `export const dynamic = 'force-dynamic'` — added because Privy's runtime check intermittently crashed static prerender in CI — so every route was already server-rendered on demand. Reading a header changes no route's mode, and the build output is identical before and after.

The parser has no dependencies and cannot throw on malformed input: a missing, empty or unrecognised UA yields `chromeMajor: null` and `belowKnownFloor: false`. An unknown engine is reported as unknown rather than old, so the number this exists to measure isn't inflated by Safari and crawlers.

`lib/logger.ts` is imported from a server component only, per `apps/web/CLAUDE.md`, and its emitter already no-ops when the OTel SDK isn't wired rather than failing the render.

**Worth a decision at review:** this logs once per page load, including the full UA string. That volume is what buys the denominator — without every request there's no percentage, only a raw count of old engines. If the OTel/PostHog volume is a concern, the alternative is to log only when `belowKnownFloor` is true and give up the ratio. I've kept the denominator; say the word if that's the wrong trade.

## Verification

- `pnpm --filter web type-check` clean.
- `pnpm --filter web test` — **408 pass**, 6 new. The parser is tested against the verbatim UA shapes of both devices from the #196 investigation (SNE-LX3 on WebView 80, Pixel 9 Pro on WebView 150), plus desktop Chrome, iOS Safari, empty/missing input, and the exact boundary at the floor constant.
- Production build green; route table byte-identical, still 21 dynamic routes and 3 static pages generated.
- Ran the production server locally and issued requests with both device UA strings. Emitted:

```
[INFO] document request {
  ua: '…SNE-LX3 Build/HUAWEISNE-L21; wv…Chrome/80.0.3987.99…',
  chromeMajor: 80,  isAndroidWebView: true,  belowKnownFloor: true
}
[INFO] document request {
  ua: '…Pixel 9 Pro; wv…Chrome/150.0.7871.181…',
  chromeMajor: 150, isAndroidWebView: true,  belowKnownFloor: false
}
```

Not verified automatically: that the records land in PostHog through the OTel exporter in a deployed environment, since the OTLP key is a server env var. Worth a look at the preview's log stream before merging.

## Scope / non-goals

Refs #196 — this measures the population, it does not fix anything. The actual fix (down-levelling dependency syntax so old engines can parse the bundle) is unsolved and blocked on the support-floor decision; the diagnosis and what was already tried are in #204.

### DIFF
--- a/apps/web/src/__tests__/lib/userAgentInsight.test.ts
+++ b/apps/web/src/__tests__/lib/userAgentInsight.test.ts
@@ -1,0 +1,62 @@
+import { describe, it, expect } from 'vitest'
+import { inspectUserAgent, KNOWN_PARSEABLE_CHROME_MAJOR } from '@/lib/userAgentInsight'
+
+// The two devices from the #196 investigation, verbatim in shape: a Huawei
+// Mate 20 Lite whose system WebView never left the 2018 factory image, and a
+// current Pixel. The first is what we cannot render on.
+const HUAWEI_WEBVIEW_80 =
+  'Mozilla/5.0 (Linux; Android 10; SNE-LX3 Build/HUAWEISNE-L21; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/80.0.3987.99 Mobile Safari/537.36'
+const PIXEL_WEBVIEW_150 =
+  'Mozilla/5.0 (Linux; Android 16; Pixel 9 Pro Build/BP41.250916.005; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/150.0.7871.181 Mobile Safari/537.36'
+const DESKTOP_CHROME_150 =
+  'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.7871.186 Safari/537.36'
+const IOS_SAFARI =
+  'Mozilla/5.0 (iPhone; CPU iPhone OS 17_5 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.5 Mobile/15E148 Safari/604.1'
+
+describe('inspectUserAgent', () => {
+  it('flags the WebView 80 device that cannot parse our bundle', () => {
+    expect(inspectUserAgent(HUAWEI_WEBVIEW_80)).toEqual({
+      chromeMajor: 80,
+      isAndroidWebView: true,
+      belowKnownFloor: true,
+    })
+  })
+
+  it('clears a current WebView', () => {
+    expect(inspectUserAgent(PIXEL_WEBVIEW_150)).toEqual({
+      chromeMajor: 150,
+      isAndroidWebView: true,
+      belowKnownFloor: false,
+    })
+  })
+
+  it('distinguishes a real browser from an embedded WebView', () => {
+    expect(inspectUserAgent(DESKTOP_CHROME_150).isAndroidWebView).toBe(false)
+    expect(inspectUserAgent(PIXEL_WEBVIEW_150).isAndroidWebView).toBe(true)
+  })
+
+  it('treats an unknown engine as unknown rather than old', () => {
+    // Safari advertises no Chrome/ token; guessing "old" here would inflate
+    // the very number this exists to measure.
+    const safari = inspectUserAgent(IOS_SAFARI)
+    expect(safari.chromeMajor).toBeNull()
+    expect(safari.belowKnownFloor).toBe(false)
+  })
+
+  it('handles a missing or empty header without throwing', () => {
+    for (const value of [null, undefined, '']) {
+      expect(inspectUserAgent(value)).toEqual({
+        chromeMajor: null,
+        isAndroidWebView: false,
+        belowKnownFloor: false,
+      })
+    }
+  })
+
+  it('puts the boundary exactly at the syntax our dependencies ship', () => {
+    const at = `Chrome/${KNOWN_PARSEABLE_CHROME_MAJOR}.0.0.0`
+    const below = `Chrome/${KNOWN_PARSEABLE_CHROME_MAJOR - 1}.0.0.0`
+    expect(inspectUserAgent(at).belowKnownFloor).toBe(false)
+    expect(inspectUserAgent(below).belowKnownFloor).toBe(true)
+  })
+})
--- a/apps/web/src/__tests__/lib/userAgentInsight.test.ts
+++ b/apps/web/src/__tests__/lib/userAgentInsight.test.ts
@@ -2,6 +2,7 @@ import { describe, it, expect } from 'vitest'
 import {
   classifyRequestKind,
   inspectUserAgent,
+  shouldRetainRawUserAgent,
   KNOWN_PARSEABLE_CHROME_MAJOR,
   SUPPORT_FLOOR_CHROME_MAJOR,
 } from '@/lib/userAgentInsight'
@@ -25,6 +26,7 @@ describe('inspectUserAgent', () => {
       isAndroidWebView: true,
       belowKnownFloor: true,
       belowSupportFloor: false,
+      isLikelyBot: false,
     })
   })
 
@@ -34,6 +36,7 @@ describe('inspectUserAgent', () => {
       isAndroidWebView: true,
       belowKnownFloor: false,
       belowSupportFloor: false,
+      isLikelyBot: false,
     })
   })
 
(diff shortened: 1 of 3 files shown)


csacanam:
Tried to close the one open verification item — that the records actually reach PostHog through the OTel exporter — and hit a permissions wall rather than a technical one.

Sent 6 requests to this PR's preview (3 with the Huawei Mate 20 Lite UA, WebView 80; 3 with the Pixel 9 Pro UA, WebView 150), then went for the runtime logs. My Vercel account only belongs to `csacanams-projects` and `peewah`; `minipay-celo-s-projects`, where `mondeto-web` lives, isn't among them, so the CLI answers `The specified scope does not exist`.

**So this still needs someone on that team to confirm.** Open the preview deployment's runtime logs and look for `document request` — the probe traffic above should show up as `chromeMajor: 80 / belowKnownFloor: true` and `chromeMajor: 150 / belowKnownFloor: false`.

What is already verified stands: the call fires with the right parsed values (checked against a local production build with both UA strings), the parser is unit-tested against those same strings, and the deployed preview returns 200 on `/`, `/faq` and `/ranks` with the WebView 80 UA — which is the part that could have broken from making the root layout `async` and reading headers.

Worth knowing about the failure mode either way: `lib/logger.ts` mirrors every record to `console.log` before attempting the OTel emit, and the emit is wrapped in a catch that no-ops. So if the exporter isn't wired in this environment, the lines still land in Vercel's log stream and nothing crashes — the risk isn't losing the data, it's having it in Vercel instead of PostHog where it can be grouped.

GigaHierz:
Reviewed. The parser is good — pure, dependency-free, correct on the boundary, and right to treat an unknown engine as unknown rather than old, which is the mistake that would have quietly inflated the number this exists to produce. Tests cover the shapes that matter. The `force-dynamic` reasoning is sound and the build-output check is the right evidence to have brought.

Two substantive problems with the **measurement**, though, before this lands. Neither is a style note — both bias the result in the direction that makes the problem look smaller than it is.

### 1. This counts requests, not users, and the bias runs the wrong way

A user whose bundle dies at parse makes **exactly one** document request and can never navigate client-side — no hydration, no router, no second render. A healthy user hydrates and then does SPA navigation; they also come back, reload, and open share links.

So healthy users contribute more log lines per person than broken users do. "Percentage of document requests below the floor" therefore **understates** "percentage of users below the floor" — which is the number this PR exists to produce, and the number the support-floor decision turns on.

There's likely a second multiplier on top: App Router may re-invoke the root layout on RSC and prefetch requests, which only healthy clients ever issue. Worth confirming rather than assuming — two-minute check on the preview:

> Load `/`, then navigate to `/profile` and `/faq` via in-app links, and count `document request` lines in the log stream. If you get more than one, RSC renders are in the denominator.

Either way the fix is the same shape: read the `RSC` / `Next-Router-Prefetch` headers and either filter to true document requests, or keep everything and emit a `requestKind` attribute so the group-by can separate them. I'd lean to the second — the RSC/document ratio is itself a decent proxy for "did this client survive hydration", which is close to the thing we actually want to measure.

### 2. It can't isolate MiniPay traffic

MiniPay is only detectable client-side, from `window.ethereum.isMiniPay` (`src/types/ethereum.d.ts`). There is no server-visible signal in the document request, and `isAndroidWebView` matches *every* embedded WebView — Facebook, Instagram, Opera's in-app browser, and assorted crawlers.

So the denominator is "all traffic", not "MiniPay traffic", and the headline percentage will be diluted by populations we don't care about. That's not a blocker — the absolute count of `belowKnownFloor` requests is still the useful number, and it's still infinitely better than what we have now. But it should be stated in the code comment, because the natural way to read the output is as a MiniPay figure, and it isn't one.

### Smaller things

- **Log volume is double what the PR describes.** `lib/logger.ts` mirrors every call to `console.log` *in addition* to the OTel emit. So this is one Vercel log line **and** one PostHog record per render. Answering your explicit question: keep the denominator, but combined with (1) it should be filtered by request kind rather than unconditional. That likely gets the volume down enough that the trade-off question goes away on its own.
- **`chromeMajor ?? -1`** — the sentinel is understandable given the primitive-only attribute constraint, but `-1` will silently poison any `avg`/`min` aggregation someone runs later. Prefer omitting the attribute entirely when it's null.
- **The full UA string is personal data** under our own privacy policy (`app/privacy/page.tsx`). I think it's fine and I'd keep it — the raw string is what lets us go back and re-parse when we find a device shape the parser mishandles. Flagging it as a conscious call rather than an oversight.
- **`KNOWN_PARSEABLE_CHROME_MAJOR`** — the support floor is now settled at Chrome 80 (see #196). Keep this constant at 85 and add a separate one for the floor: 85 is *what our bundle can parse today*, 80 is *what we intend to support*. The gap between them is the remaining work, and collapsing them into one number hides it.
- Minor conflict with #204 — both add imports to `layout.tsx`. Whichever lands second rebases.

Happy to merge once (1) is addressed. (2) just needs a comment saying so.

GigaHierz:
Second review pass, deliberately aimed at what the thread above doesn't already cover:
parser correctness against real UA strings, ReDoS on the attacker-controlled header,
bot traffic in the denominator — and the RSC/prefetch question, which I went and
settled against the Next source rather than leaving as a suspicion. **One point above
turns out not to hold**, and I think it changes the requested fix.

### Strengths

- **The measurement premise is sound and the placement is right.** `layout.tsx:59`
  does already carry `export const dynamic = 'force-dynamic'` (pre-existing,
  unchanged in this diff), so "no route's rendering mode changes" is accurate. `next`
  is pinned `^15.5.20`, where `headers()` is async — `(await headers())` at
  `layout.tsx:75` is correct, and making `RootLayout` async has no other consequence
  (no `loading.tsx`, no PPR, no server actions — all verified).
- **"Cannot throw on malformed input" holds.** There is no throw path in
  `userAgentInsight.ts:35-49`: `??`, `RegExp.exec`, `Number`, `<`. Checked against 17
  real and adversarial shapes.
- **The `CriOS` trap is avoided** — and this is the one most UA parsers get wrong.
  Chrome-for-iOS reports `CriOS/126.…` with no `Chrome/` token, so `/Chrome\/(\d+)/`
  correctly yields `null`. Same for `EdgiOS/`. Had it matched, every iOS Chrome user
  would have been bucketed with a WebKit major that means nothing.
- **The `Edg/` / `OPR/` reasoning at `userAgentInsight.ts:38-40` is correct, verified:**
  Edge desktop → 120, Opera Android → 86, Samsung Internet 10.1 → 71. Taking the
  Chromium base rather than the vendor's numbering is right for a *parse-capability*
  question, and the comment explains why.
- **`85` is the right number for the right reason.** `&&=`/`||=`/`??=` shipped in V8
  8.5 / Chrome 85, the boundary is on the correct side (`85 < 85` is false, so 85 is
  *not* below the floor), and the test at `userAgentInsight.test.ts:53-58` asserts
  exactly that side.
- The new module lands under `src/lib/**`, the one path with a real coverage floor,
  and arrives fully covered.

### Issues

#### Critical
None. The parser is sound and the wiring is correct; nothing here can break a render.

#### Important

**1. The root layout does *not* re-render on RSC navigations or prefetches — so the
suspected second multiplier doesn't exist in this app, and `requestKind` doesn't fix
the bias it was proposed to fix.**

Checked in the Next 15.5 source rather than reasoned from first principles. An RSC
request goes `app-render.js:1069` → `generateDynamicFlightRenderResult` →
`generateDynamicRSCPayload` (`app-render.js:196`) → `walkTreeWithFlightRouterState`,
where:

```js
const renderComponentsOnThisLevel =
  !flightRouterState ||                                   // client sent no state tree
  !matchSegment(actualSegment, flightRouterState[0]) ||   // segment mismatch
  parallelRoutesKeys.length === 0 ||                      // leaf
  flightRouterState[3] === 'refetch';                     // explicit refresh
```

`createComponentTree` — the only thing that invokes the layout component — is called
only when that is true. At the root level of a soft navigation all four are false: the
router always sends `next-router-state-tree`, the root segment matches itself, the
root has children, and the state is not `refetch`. So the root layout is never
entered and no `document request` line is emitted.

Prefetches are exempt even more clearly: `walk-tree-with-flight-router-state.js:61-64`
short-circuits to router-state-only when `isPrefetch && !modules.loading &&
!hasLoadingComponentInTree(...)`, and this app has **no `loading.tsx` anywhere**. PPR
is off, so `shouldProvideFlightRouterState` is true for prefetches too and the same
segment-match logic applies.

The only two remaining ways to re-enter the root layout are also absent:
`flightRouterState[3] === 'refetch'` (set by `router.refresh()` — zero occurrences in
`src/`) and server actions (zero `'use server'` in `src/`).

**Consequence:** healthy users do not generate extra log lines by navigating. Every
internal navigation here is a soft navigation, so a healthy session is ~1 document
request — the same as a broken one. The residual bias is not an SPA-navigation
multiplier; it's a *sessions-per-person over the window* effect, where a retained
healthy user returns on later days but a frustrated user may also reload 2–3 times
before leaving. Those pull in opposite directions and the net sign is genuinely
unknown rather than known to understate.

That matters because filtering on `RSC` / `Next-Router-Prefetch` will today classify
~100% of records as `document` and correct nothing. I'd still add the field — as
**cheap insurance**, not as the fix. The moment someone adds a `loading.tsx`, calls
`router.refresh()`, or enables PPR, the assumption breaks silently with no test to
catch it. Headers are lowercase and copy-pasteable
(`app-router-headers.js:83-90`):

```ts
const h = await headers()
const requestKind =
  h.get('next-router-prefetch') ? 'prefetch'
  : h.get('rsc') ? 'rsc'
  : 'document'
```

*If the goal is actually "what fraction of clients never boot", this PR unlocks a
cleaner measurement with none of this bias:* the ratio of server-side `document
request` records to client-side `app_opened` events. `app_opened` fires once per
browser session (`posthog-provider.tsx:11-23`) and **only from a client that survived
parse and hydrated**. `document_requests / app_opened` is the boot-failure rate
directly — no UA parsing, no per-session dedupe problem. Worth building the dashboard
on both numbers.

**2. Bot and crawler traffic lands in both the numerator and the denominator, and
nothing filters it.** Two separate effects, verified by running the parser:

- *Numerator poisoning.* Pinned-old-Chrome UAs are ubiquitous among scrapers and
  uptime checks. `Chrome/41.0.2228.0` (the classic old-Googlebot spoof, still shipped
  by a lot of tooling) → `{chromeMajor: 41, isAndroidWebView: false, belowKnownFloor:
  true}`. So does `UCBrowser` at `Chrome/57`. Indistinguishable in the log from a real
  broken device.
- *Denominator inflation.* The public share route (`src/app/s/page.tsx`) gets unfurled
  by Twitterbot, `facebookexternalhit`, WhatsApp, Slack and Discord on every share.
  Those render a page, so they log — and mostly carry no `Chrome/` token, inflating
  the denominator with `chromeMajor: null`. Given the share flywheel, the non-human
  share of the denominator is not marginal. (`/api/og` is a route handler with no
  layout, so it's correctly not logged.)

Cheap fix, keeping the record so the denominator stays correctable: add an
`isLikelyBot` boolean from one regex — `/bot|crawl|spider|slurp|preview|monitor|headless|lighthouse|facebookexternalhit|whatsapp|telegram|Applebot|Yandex/i`
— and report the headline ratio over `chromeMajor !== null` only, gated on
`isAndroidWebView && belowKnownFloor`. The desktop-old-Chrome bucket is essentially
all bots and should never reach the number quoted in #196. (Next also exports
`userAgent()` from `next/server` with an `isBot` field if you'd rather not own the
regex.)

**3. The floor constant is a snapshot with nothing enforcing it, in a repo where
Renovate bumps dependencies continuously.** `KNOWN_PARSEABLE_CHROME_MAJOR = 85`
(`userAgentInsight.ts:24`) was derived from one investigation of one bundle. `||=`/`??=`
is the *lowest* new syntax found, not necessarily the highest — class static blocks
are Chrome 94, top-level `await` 89, `Object.hasOwn` 93. The next dependency bump can
raise the real parse floor above 85 while this constant stays put, silently turning
`belowKnownFloor: false` into a false negative for exactly the population being
measured. At minimum record *how* 85 was derived so it can be re-derived; better, add
a build-step syntax check over `.next/static/chunks` so the constant is asserted
rather than remembered.

#### Minor

**4. `; wv)` misses pre-Lollipop WebViews — the oldest tail, which is what this hunts.**
`userAgentInsight.ts:46`. The `wv` token arrived with the Android 5.0 standalone
WebView; an Android 4.4 WebView reports `… Version/4.0 Chrome/30.0.0.0 Mobile
Safari/537.36` with no `wv`. Verified: that string yields `{chromeMajor: 30,
isAndroidWebView: false, belowKnownFloor: true}` — still in the numerator, but it drops
out if the analysis is gated on `isAndroidWebView && belowKnownFloor` as recommended
in #2. The `Version/\d` + `Chrome/` pair is the classic WebView tell that survives back
to 4.4. Bounded blast radius given MiniPay's minimum Android, so genuinely minor.

**5. The metric is attacker-writable.** `curl -H 'User-Agent: Chrome/1'` produces
`belowKnownFloor: true`. Not a security issue — nothing branches on it — but the
aggregate should be sanity-checked against the raw UA distribution before it's quoted
in a support-floor decision, rather than read straight off a dashboard tile.

**6. Absurd version strings are handled.** `Chrome/99999999999999999999` → `1e20`, not
below floor, no throw. `Chrome/0080` → `80`. Googlebot's literal `Chrome/W.X.Y.Z` →
`null`. All fine, just untested.

**7. `logger.info` is the first unconditional call site in the codebase.** Every
existing `logger.*` call in `src/app/api/**` is a `warn`/`error` on a failure path.
This turns an exception channel into a per-request stream — worth stating in the body
as a deliberate posture change, so nobody later reads a `document request` line as a
sign something went wrong.

**8. Missing test cases, named.** Six solid tests, but they pin only the happy path.
Not covered, all of which are behaviours the source comments explicitly claim:
`CriOS` and `EdgiOS`; `Edg/` and `OPR/` (`userAgentInsight.ts:38-40` states a
deliberate behaviour that no test pins, so a future "improvement" could break it
silently); Samsung Internet on an old Chromium; an Android WebView UA *without* `; wv`;
a spoofed old-Chrome bot string; `Chrome/W.X.Y.Z`; an absurd version. The boundary test
also uses the bare fragment `Chrome/85.0.0.0` rather than a full UA, so no test
exercises the boundary in a realistic string.

**9. ReDoS: checked, and there is none.** Recording it so nobody re-raises it. Both
patterns are linear — `/Chrome\/(\d+)/` has no nested quantifier, and `/;\s*wv\)/` can
only enter `\s*` after a literal `;`, so total backtracking is bounded by the sum of
whitespace-run lengths, i.e. O(n). Measured on 16 KB adversarial inputs (8k `;` + 8k
spaces; `"; "` × 4000; `Chrome/` + 16k digits; `Chrome/` × 2000): **worst case 0.053 ms
per call**. Node's 16 KB header cap keeps this permanently negligible.

### Recommendations

**Direct answer to the volume question — keep the denominator, but neither option as
posed.** The binary gives up more than it needs to:

1. **Keep one record per document request.** You need the denominator, and per #1 it's
   much cleaner than believed.
2. **Log the raw `ua` string conditionally** — only when `chromeMajor === null ||
   belowKnownFloor || isAndroidWebView`. Everything else gets the three parsed fields
   and no UA. Same exact denominator, ~70% fewer bytes for the healthy majority, and
   the retained personal-data surface shrinks to precisely the population under
   investigation. This dominates both options in the body.
3. **Time-box it.** This is a census answering one question. Say in the body when it
   comes out — a fixed window, then delete or flip to `belowKnownFloor`-only — and open
   the follow-up now, or it becomes permanent by default.
4. **Add `requestKind`** per #1, as a regression tripwire, with a comment saying it's
   expected to read `document` ~100% of the time today and why.
5. **Add `isLikelyBot`** per #2, and define the headline number as `isAndroidWebView &&
   belowKnownFloor && !isLikelyBot` over `chromeMajor !== null`.
6. **Now the floor is settled at 80, a boolean is the wrong shape.** The number #196
   needs is a three-way bucket: `< 80` (out of scope), `80 ≤ major < 85` (**intended to
   support, currently broken — this is the band that sizes the remaining work**), `≥ 85`
   (fine). A single `belowKnownFloor` collapses the first two, which are the two that
   must be told apart.
7. **On the stdout mirror** (`lib/logger.ts:32-39`): every document request now also
   writes to Vercel's runtime log stream, which is volume-limited — during an incident
   this can push the `error` lines you need out of the retention window. Either accept
   it knowingly, or give `logger` a way to skip the console mirror for high-volume
   records.
8. **Fix the `-1` sentinel** at `layout.tsx:79`. Beyond poisoning aggregations, it
   undoes the PR's own headline guarantee: the parser is careful that "unknown ≠ old",
   then the logging boundary encodes unknown as a number *lower than every real
   version*, so any `chromeMajor < 85` filter in PostHog counts every Safari and every
   crawler as ancient. `Attributes` in `lib/logger.ts:22` accepts `undefined`, so
   omitting the key is available today. Note the coverage gate can't catch this —
   `vitest.config.ts` excludes `src/app/**/layout.tsx`.

### Assessment

**Ready to merge?** With fixes — but smaller ones than the thread currently implies.

**Reasoning:** The parser is genuinely correct — `CriOS`, `Edg/`, `OPR/`,
`SamsungBrowser/` and malformed input all behave as documented, `85` is right for the
right reason, and the regexes are ReDoS-free at 0.05 ms on 16 KB adversarial input.
What actually needs fixing before merge is the `-1` sentinel (it contradicts the PR's
own stated guarantee) and some form of bot handling (unaddressed, and material given
the share-unfurl traffic on `/s`). The request-kind concern currently gating the merge
is, per the Next 15.5 source, largely not real in this app and is better added as a
tripwire than treated as a correctness blocker.

---

*One caveat worth stating plainly: finding #1 is high but not total confidence (~90%).
I read the relevant Next internals at 15.5.19 against this repo's pinned `^15.5.20` —
same minor, one patch apart — and verified the three escape hatches are absent by
grep, but I did not run it end to end. It's falsifiable in 30 seconds on the preview:
click an internal `<Link>` and see whether a second `document request` line appears. If
one does, the original reading stands and mine is wrong.*

csacanam:
Addressed the review. Pushed as a single commit.

**(1) Counting requests rather than users — fixed.** Emitting `requestKind` (`document` / `rsc` / `prefetch`) rather than filtering, which was the option you leaned to. The document-only slice gives the per-visit denominator, and the RSC-to-document ratio stays available as the hydration-survival proxy you pointed at. `classifyRequestKind` is pure and tested, including the ordering trap: Next sends both headers on a prefetch, so testing `RSC` first would bury every prefetch in the `rsc` bucket.

Your two-minute check turned out to be unnecessary to run — the fix doesn't depend on the answer either way, since both readings survive.

**(2) Can't isolate MiniPay — recorded in the code**, as you asked, including why (`window.ethereum.isMiniPay` is client-only, and `isAndroidWebView` matches Facebook, Instagram, Opera and crawlers alike) and the explicit warning not to read the percentage as a MiniPay figure.

**Support floor split out.** `KNOWN_PARSEABLE_CHROME_MAJOR` stays at 85; `SUPPORT_FLOOR_CHROME_MAJOR` is 80 per #196. The insight now reports both, which splits the census three ways: below 80 is out of scope by decision, **80–84 is in scope and broken**, 85+ is fine. That middle bucket is the remaining work, and it's the one that disappears if the two numbers are collapsed. There's a test asserting the gap exists rather than just the boundary.

**`chromeMajor ?? -1` — gone.** The attribute is omitted entirely when the UA advertises no Chromium version, so no `avg`/`min` gets poisoned later.

**Log volume.** Kept unconditional, per your answer to keep the denominator. With `requestKind` on every record the volume is now filterable at query time rather than at emit time, so the trade-off you flagged resolves itself.

Verification: `tsc --noEmit` clean, **412 tests pass** across 50 files (10 in `userAgentInsight.test.ts`, up from 6).

Still unverified, same as before: that the records actually arrive in PostHog through the OTel exporter. That needs someone on `minipay-celo-s-projects` — my Vercel account isn't in that scope. Worth knowing it's a low-stakes unknown: `lib/logger.ts` mirrors every record to `console.log` before the OTel emit and swallows exporter failures, so the data lands in Vercel's log stream regardless; the risk is having it in the wrong place, not losing it.

Also worth flagging: #221 is a live browser-only outage on `main`, unrelated to this branch but it will affect anything you load to test. MiniPay is unaffected.

csacanam:
Correction to my comment above: I answered the first review and hadn't read the second pass when I wrote it. Two things in it were wrong.

**I said your two-minute check "turned out to be unnecessary to run".** You had already settled it — against the Next 15.5 source — and found that the root layout isn't re-entered on soft navigations or prefetches, so `requestKind` doesn't correct the bias it was proposed for. I asserted the opposite of your finding while claiming to have addressed it. `classifyRequestKind`'s docblock now carries your reasoning and says plainly that it should read `document` for essentially every record today, and that it exists for the moment someone adds a `loading.tsx`, calls `router.refresh()` or turns on PPR.

**I also claimed the volume trade-off "resolves itself".** It doesn't, and your recommendation 2 is better than either option I'd posed: the raw `ua` now rides along only when `chromeMajor === null || belowKnownFloor || isAndroidWebView`. Same denominator, and the retained personal data shrinks to exactly the population under investigation.

## Now addressed

- **`isLikelyBot`** (#2), with your regex. Tested against both failure modes you named: the `Chrome/41.0.2228.0` Googlebot spoof, which parses as a real broken engine, and the unfurler set — `facebookexternalhit`, Twitterbot, WhatsApp, Telegram, HeadlessChrome. The headline definition is recorded in `layout.tsx`: `isAndroidWebView && belowKnownFloor && !isLikelyBot` over records with a known major.
- **Three-way bucket** (#6). `SUPPORT_FLOOR_CHROME_MAJOR = 80` alongside the parse floor, so 80–84 — intended to support, currently broken — is separable from out-of-scope. A test asserts the gap exists, not just the boundary.
- **`-1` sentinel** (#8), omitted entirely. Your framing was the sharper one: it undid the parser's own unknown-is-not-old guarantee.
- **How `85` was derived** (#3), recorded on the constant, including that `||=`/`??=` was the lowest new syntax found rather than provably the highest, and that Renovate can raise the real floor underneath it.
- **The `logger.info` posture** (#7), stated where it happens — including that the stdout mirror can push `error` lines out of Vercel's retention during an incident. Accepted knowingly, which is one of the two options you offered.
- **The named test gaps** (#8 under Issues). `CriOS`, `EdgiOS`, `Edg/`, `OPR/`, `SamsungBrowser`, the Android 4.4 WebView with no `wv`, the spoofed bot, `Chrome/W.X.Y.Z`, and the absurd version. The boundary test now uses a full UA string rather than a bare fragment. 19 tests in this file, up from 6.

`tsc --noEmit` clean; **421 tests pass** across 50 files.

## Not addressed, deliberately

- **Time-boxing** (recommendation 3). You're right that it becomes permanent by default, but picking the window is a call I'd rather not make alone — say when you want the number and I'll open the follow-up with that date in it.
- **The build-step syntax check** over `.next/static/chunks` (#3). I took the "at minimum" option and recorded the derivation. The asserted version is the right end state and is its own piece of work.
- **`document_requests / app_opened`** — noted and I agree it's the cleaner measurement, but it's a dashboard decision rather than a code change, so nothing in this diff commits us either way.

One caveat on your caveat: I didn't independently re-verify the Next internals, so finding #1 is taken on your reading. The 30-second falsification you describe still stands as the cheapest way to settle it, and `requestKind` is now what would surface it if you're wrong.

GigaHierz (approved):

PR #187, opened 2026-08-04
feat: live on-map purchase activity feed

GigaHierz:
## What

Adds a live purchase feed to the map: a small toast slides into the bottom-left corner when someone buys pixels on the **current map** — "someone just bought N pixels for \$X" — then auto-dismisses. A real, verifiable social-proof / FOMO signal (every purchase is an on-chain event), not a fabricated one.

## How

Frontend-only — it rides the **existing** shipped Goldsky subgraph (`apps/subgraph`); **no new indexer and no new env var**.

- **`lib/subgraph.ts`** — `fetchRecentBatches` + `fetchProfilesFor` (query `purchaseBatch` + `ownerProfile`), matching the existing `querySubgraph` conventions.
- **`api/activity`** — cached (~8s) route, **gated on `subgraphConfigured()`** so it returns an empty feed as a fallback (same pattern as `/api/pnl`, `/api/analytics`); decodes the profile label via `lib/decodeBytes`.
- **`useActivityFeed`** — polls every 10s, **seeds on mount** (never floods history), dedupes by batch id, drains **one toast at a time**, and **excludes the viewer's own buys** (they already get tx-receipt feedback).
- **`ActivityToast`** — self-contained corner toast: auto-dismiss, pause-on-hover, dismiss-on-tap; `activityIn` entrance keyframe; mounted in `page.tsx`.

Buyer names come from the on-chain profile label, falling back to the deterministic `generateUsername`; money is already 6-dec microcents (`formatUSDT`-ready). Uses the same `NEXT_PUBLIC_GOLDSKY_SUBGRAPH_URL` as the leaderboard/analytics routes.

## Verification

- `tsc` clean · **409 tests pass** (9 new for the feed's dedupe/seed/own-exclude/format logic) · production build green (`/api/activity` in the route table).
- Ran `/api/activity` locally against the live `mondeto/prod` subgraph: real batches, on-chain labels decoded correctly (e.g. `Victor-grosser`), `null` label where no profile.

## Deploy note

No new env var. If `NEXT_PUBLIC_GOLDSKY_SUBGRAPH_URL` is set on **Production** (as the leaderboard/analytics already require), the feed activates on deploy; otherwise it stays quiet via the fallback.

Closes #188

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #209, opened 2026-08-10
perf(minipay): stop shipping Privy, x402 and @solana/kit to clients that never use them

### LINKED ISSUE #196: MiniPay freezes and force-closes on entering Mondeto — player cannot buy at all

### LINKED ISSUE #208: Privy + x402 + @solana/kit ship to MiniPay users who can never use them
## Summary

MiniPay users download, parse and execute the entire Privy + x402 + `@solana/kit` dependency graph, then render a placeholder button. They can never use any of it — MiniPay surfaces an injected wallet and never mounts the Privy tree.

`wallet-provider.tsx` is explicitly designed to prevent this. Its architecture comment says:

> MiniPay users keep the vanilla tree and the injected connector auto-connects. The Privy SDK never loads for them.

That holds at runtime. It does not hold at bundle time, because `connect-button.tsx` reaches `@privy-io/react-auth` through **two** static imports that webpack resolves at build time regardless of any runtime guard.

## The two leaks

`apps/web/src/components/connect-button.tsx`:

```ts
import { useConnectWallet } from "@privy-io/react-auth";   // L3  — direct
import { PrivyReadyContext } from "./wallet-provider-privy"; // L10 — indirect
```

The second is the easier one to miss: it looks like it imports a bare `createContext(false)`, but `wallet-provider-privy.tsx` statically imports `PrivyProvider` from `@privy-io/react-auth` and `createConfig`/`WagmiProvider` from `@privy-io/wagmi` at module scope. Pulling one named export in pulls the whole module graph.

`ConnectButton` is mounted from `components/Layout/TopBar.tsx` and `components/navbar.tsx`, so it is on every page — including the map, which is the MiniPay entry path.

Resulting graph on a MiniPay load:

```
TopBar → connect-button.tsx
           → @privy-io/react-auth            (L3 and L10)
               → x402@0.7.3                  (non-optional dep of Privy)
                   → @solana/kit, @solana/offchain-messages,
                     @solana/transactions, @solana-program/*
               → @base-org/account → @coinbase/cdp-sdk → @solana/web3.js
```

The `next/dynamic({ ssr: false })` split on `wallet-provider-privy.tsx` in `wallet-provider.tsx:58` is correct and does its job — it's just bypassed by a second, non-lazy path to the same module.

`PrivyReadyContext` is doing its job too. It correctly stops `useConnectWallet` from being called outside its provider. But it's a **runtime** guard, and this is a **bundling** problem: webpack has no way to know the context will be `false`.

## Why it matters

- It is a large download, parse and execute cost imposed on the lowest-end devices we serve, for code those devices can never reach.
- It is entangled with #196. `@solana/offchain-messages` and `@solana/transactions` ship `||=`, which is Chrome 85+, and the snippet identified as throwing on the WebView 80 device (`(u ||= {})`) traces to exactly there. Fixing this leak removes the Solana tree from the MiniPay path and shrinks the scope of the down-levelling work.

It does **not** close #196. With the leak fixed, `viem` (12 files), `@noble/hashes` (6) and `@wagmi/connectors` (1) still ship `||=`/`??=` on the MiniPay path, and all three are genuinely required. The transpile pass is still needed — it just no longer has to cover Solana.

## Proposed fix

The component already has the runtime boundary in the right place:

```ts
if (!privyReady) return <ConnectButtonPlaceholder />;
return <ConnectButtonInteractive />;
```

`ConnectButtonInteractive` is already a separate function and the only consumer of `useConnectWallet`. So:

1. **Move `PrivyReadyContext` into its own module** (e.g. `components/privy-ready-context.ts`) containing nothing but `createContext(false)`. `wallet-provider-privy.tsx` and `connect-button.tsx` both import it from there. Kills the L10 leak.
2. **Move `ConnectButtonInteractive` into its own module** and load it with `next/dynamic({ ssr: false })`, using `ConnectButtonPlaceholder` as the `loading` fallback so there's no layout shift. Kills the L3 leak.

After that, `wallet-provider-privy.tsx` is the only module referencing `@privy-io/*`, and it is already lazy.

## Acceptance

A production build in which **no chunk reachable on a MiniPay load** (`window.ethereum.isMiniPay === true`) contains `@privy-io`, `x402`, or `@solana` code. Verified against built output, not by inspection — a static import is exactly the kind of thing that reads as fine and bundles as not.

Refs #196.

GigaHierz:
## Problem

MiniPay clients download, parse and execute the entire Privy dependency graph — including `x402` and `@solana/kit` — and then render a placeholder button. They never mount `PrivyTree`: MiniPay surfaces an injected wallet, `PrivyReadyContext` stays `false`, and none of that code is reachable for them.

`wallet-provider.tsx` is built to prevent exactly this, and its architecture comment says so:

> MiniPay users keep the vanilla tree and the injected connector auto-connects. The Privy SDK never loads for them.

True at runtime. Not true at bundle time, because `connect-button.tsx` — which `TopBar` and `navbar` mount on **every page, including the map** — reached `@privy-io/react-auth` through two static imports:

```ts
import { useConnectWallet } from "@privy-io/react-auth";     // direct
import { PrivyReadyContext } from "./wallet-provider-privy";  // indirect
```

The second is the one that hides well. It looks like it imports a bare `createContext(false)`, but `wallet-provider-privy.tsx` imports `PrivyProvider` and `@privy-io/wagmi` at module scope, so pulling one named export in pulled the whole graph.

`PrivyReadyContext` was doing its job — it correctly stops `useConnectWallet` being called outside its provider. But it is a *runtime* guard against what is a *bundling* problem. Webpack resolves static imports at build time and has no way to know the context will be `false`.

## What changed

1. **`privy-ready-context.ts`** (new) — `PrivyReadyContext` moved into a module that imports nothing but React. Both `wallet-provider-privy.tsx` and `connect-button.tsx` read it from there.
2. **`connect-button-interactive.tsx`** (new) — the Privy-dependent half of `ConnectButton`, loaded via `next/dynamic({ ssr: false })` with the existing `ConnectButtonPlaceholder` as its `loading` fallback, so there is no layout shift.
3. **`connect-button-styles.ts`** (new) — the shared `buttonClassName` / `buttonStyle`, so the placeholder and the real button cannot drift apart. The placeholder *is* the loading state for the real one, so drift would show as layout shift on the slowest connections.

The split lands exactly where the runtime boundary already was:

```ts
if (!privyReady) return <ConnectButtonPlaceholder />;
return <ConnectButtonInteractive />;
```

`wallet-provider-privy.tsx` is now the only module referencing `@privy-io/*`, and it was already lazy. No behaviour change for anyone — desktop users get the identical Privy flow, one chunk later.

## Measured

Same machine, production builds before and after, counting the JS chunks the `/` route actually loads per `app-build-manifest.json`:

| | before | after |
|---|---|---|
| First Load JS on `/` | 860 kB | **339 kB** |
| Raw chunk bytes for `/` | 3043 kB | **1304 kB** |
| `\|\|=` / `??=` occurrences | 21 | **6** |
| Chunks carrying Privy / x402 / Solana | 3 | **0** |

Before, the offenders on `/` were `286-cd92f5e4412c85c4.js` (1307 kB, 14 operators, `@solana` + `x402` + `privy`) and `d0466ee9-a7ab8f43b9ddc0ff.js` (396 kB, `privy`) — **the same two chunk names reported as throwing on the WebView 80 device in #196**. Both are gone from the route.

Acceptance from #208 met: no chunk on `/`, `/profile`, `/atlas` or `/ranks` contains `@privy-io`, `x402` or `@solana` code. Privy still builds as its own lazy chunks for the non-MiniPay path.

## Relationship to #196

This removes the Solana tree from the MiniPay path, and the exact snippet identified as throwing there — `(u ||= {})` — traces to `@solana/offchain-messages`. So it very likely fixes the reported symptom.

**It does not close #196, and shouldn't be treated as the fix.** The remaining 6 operators come from `ox`/`viem` and friends: `viem` (12 files), `@noble/hashes` (6) and `@wagmi/connectors` (1) still ship syntax above the Chrome 80 support floor, and all three are genuinely required. The dependency down-levelling work is still needed — it just no longer has to cover Solana. Confirming on a real WebView 80 device is the only way to know whether the map now draws.

## Regression guard

`__tests__/components/minipay-privy-isolation.test.ts` asserts at source level that none of the MiniPay-path modules statically imports `@privy-io/*` or `wallet-provider-privy`, that the interactive half is still `next/dynamic`'d, and that `privy-ready-context.ts` imports only React.

Source level rather than built output on purpose: it needs to fail on the PR that reintroduces the import, not after a production build. Verified it actually fails — reinstating the `wallet-provider-privy` import turns it red, and only it.

The subtle case is covered too: the regression that caused this wasn't an obvious Privy import, it was importing a context from a module that happened to import Privy.

## Verification

- `pnpm --filter web type-check` — clean.
- `pnpm --filter web test` — **416 pass**, 9 new.
- Production build green; route table unchanged, still 21 dynamic routes and 3 static pages.
- Chunk contents inspected via `app-build-manifest.json` before and after, numbers above.

Not verified: behaviour on a real old-WebView device, and the desktop Privy connect flow end-to-end in a browser — worth a click-through on the preview before merge, since there is no existing test coverage for `ConnectButton`.

Note: `pnpm --filter web lint` fails on this branch and on `main` alike — `next lint` is deprecated and drops into an interactive ESLint setup prompt. Pre-existing, unrelated, and not what CI gates on.

Closes #208. Refs #196.

### DIFF
--- a/apps/web/src/__tests__/components/minipay-privy-isolation.test.ts
+++ b/apps/web/src/__tests__/components/minipay-privy-isolation.test.ts
@@ -1,0 +1,90 @@
+import { describe, it, expect } from 'vitest'
+import fs from 'node:fs'
+import path from 'node:path'
+
+/**
+ * Guards the module boundary that keeps `@privy-io/*` — and the `x402` /
+ * `@solana/kit` subtree it drags with it — out of the chunk MiniPay clients
+ * download.
+ *
+ * MiniPay renders with the device's Android System WebView, which on the
+ * phones this matters for is a 2018 factory build. Those clients get the
+ * injected connector and never mount `PrivyTree`, so they can never use any
+ * of that code — but webpack resolves static imports at build time, and one
+ * named import from a module mounted by `TopBar` is enough to put the whole
+ * graph on every page.
+ *
+ * This is asserted at source level rather than against built output because
+ * it needs to fail in CI on the PR that reintroduces the import, not after a
+ * production build. The realistic regression is someone adding `import { X }
+ * from '@privy-io/react-auth'` back to the shell — or, more subtly, importing
+ * something from a module that itself imports Privy, which is exactly how
+ * this happened the first time (`PrivyReadyContext` used to live in
+ * `wallet-provider-privy.tsx`).
+ */
+
+const COMPONENTS = path.join(process.cwd(), 'src/components')
+
+function read(file: string): string {
+  return fs.readFileSync(path.join(COMPONENTS, file), 'utf8')
+}
+
+/** Static `import ... from '<spec>'` sources only — ignores `import()`. */
+function staticImportSources(source: string): string[] {
+  return Array.from(
+    source.matchAll(/^\s*import\s[^;]*?from\s+['"]([^'"]+)['"]/gm),
+    (m) => m[1],
+  )
+}
+
+// Modules reachable from `TopBar` / `navbar` on every page, including the map,
+// which is the MiniPay entry path. None of these may reach Privy.
+const MINIPAY_PATH_MODULES = [
+  'connect-button.tsx',
+  'connect-button-styles.ts',
+  'privy-ready-context.ts',
+]
+
+describe('MiniPay bundle isolation', () => {
+  it.each(MINIPAY_PATH_MODULES)(
+    '%s does not statically import @privy-io',
+    (file) => {
+      const privy = staticImportSources(read(file)).filter((s) =>
+        s.startsWith('@privy-io'),
+      )
+      expect(privy).toEqual([])
+    },
+  )
+
+  it.each(MINIPAY_PATH_MODULES)(
+    '%s does not statically import wallet-provider-privy, which imports Privy',
+    (file) => {
+      const viaProvider = staticImportSources(read(file)).filter((s) =>
+        s.includes('wallet-provider-privy'),
+      )
+      expect(viaProvider).toEqual([])
+    },
+  )
+
+  it('the Privy-dependent half is loaded with next/dynamic', () => {
+    const shell = read('connect-button.tsx')
+    // If this is ever changed to a static import the assertions above still
+    // fail, but this pins the mechanism so the intent survives a refactor.
+    expect(shell).toMatch(
+      /dynamic\(\s*\(\)\s*=>\s*import\(['"]\.\/connect-button-interactive['"]\)/,
+    )
+    expect(shell).toMatch(/ssr:\s*false/)
+  })
+
+  it('the interactive half is the only place Privy is reached from', () => {
+    // Sanity check that the split actually moved the dependency rather than
+    // dropping the feature: something still has to import the Privy hook.
+    const interactive = read('connect-button-interactive.tsx')
+    expect(staticImportSources(interactive)).toContain('@privy-io/react-auth')
+  })
+
+  it('PrivyReadyContext lives in a module with no Privy dependency', () => {
+    const context = read('privy-ready-context.ts')
+    expect(staticImportSources(context)).toEqual(['react'])
+  })
+})
--- a/apps/web/src/components/connect-button-interactive.tsx
+++ b/apps/web/src/components/connect-button-interactive.tsx
@@ -1,0 +1,116 @@
+"use client";
+
+import { useConnectWallet } from "@privy-io/react-auth";
+import { useAccount, useDisconnect } from "wagmi";
+import { useEffect, useRef, useState } from "react";
+import Link from "next/link";
+import { generateUsername } from "@/lib/username";
+import { useProfile } from "@/hooks/useProfile";
+import { useMaps } from "@/hooks/useMaps";
+import { buttonClassName, buttonStyle } from "./connect-button-styles";
+
+/**
+ * The half of ConnectButton that talks to Privy.
+ *
+ * Split into its own module purely for bundling. `connect-button.tsx` is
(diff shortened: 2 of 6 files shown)


GigaHierz:
Two updates.

**Desktop Privy flow verified.** The click-through on the preview works — CONNECT opens the Privy modal and the connected menu behaves as before. That was the one behaviour this change could plausibly have broken and the one thing with no test coverage, so it's now covered by hand.

**Rebased onto `main` after #187 landed.** That PR mounts `ActivityToast` in `page.tsx`, i.e. on the map, which is the MiniPay entry path — so the numbers in the description were measured against a `/` route that no longer exists. Re-measured on the rebased branch rather than assuming they carried over:

| | before | after (original) | after (rebased onto #187) |
|---|---|---|---|
| First Load JS on `/` | 860 kB | 339 kB | **340 kB** |
| Raw chunk bytes for `/` | 3043 kB | 1304 kB | **1307 kB** |
| `\|\|=` / `??=` occurrences | 21 | 6 | **6** |
| Chunks carrying Privy / x402 / Solana | 3 | 0 | **0** |

The activity toast costs about 1 kB on the route and pulls nothing into the MiniPay path. The result holds.

`type-check` clean, **425 tests pass** (up from 416 — #187 brought 9 of its own).

Issue #199, opened 2026-08-05
chore(ci): CI checks + persistent staging environment for team collaboration


GigaHierz:
## Why

Handing the repo to a growing team. Today CI is a single `web` type-check + test job on `main`, staging is half-wired (branch exists, URL is a `<TODO>`, competing branches), and branch protection lives only as prose in docs. This makes the collaboration flow explicit and enforced — basic industry standard, nothing fancy, solid enough for ~3 people.

Current CI: `.github/workflows/ci.yml`, one `web` job — `pnpm install --frozen-lockfile` → `pnpm --filter web run type-check` → `... run test` (Vitest, 49 files), triggered only on `main`. No lint, no build gate, no subgraph/contracts coverage.

## A. Expand CI (`.github/workflows/ci.yml`)

Keep the single `web` job; add to it:

- [ ] `pnpm --filter web run lint` (`next lint` — currently review-only)
- [ ] `pnpm --filter web run build` (`next build` — catch build breaks before Vercel does)
- [ ] Run on PRs targeting **both `main` and `staging`** (today it's `branches: [main]` only)
- [ ] (optional) add `.nvmrc` = `20` so local dev matches CI; tidy the loose `engines` (root says 18+, subgraph 20+, CI hardcodes 20)

Out of scope here: no subgraph/contracts CI (separate follow-up if wanted).

## B. Persistent staging environment (Vercel)

Flow: **feature branch → PR into `staging` → verify on staging URL → merge `staging` → `main` → production.**

- [ ] Give the `staging` branch a stable Vercel domain (branch→domain assignment or a dedicated staging alias)
- [ ] Set staging-scoped env vars in Vercel; keep `/dev/*` exposed on non-prod via the existing `VERCEL_ENV` gating (`apps/web/src/app/dev/layout.tsx`)
- [ ] Fill the placeholder staging URL in `docs/MINIPAY_SUBMISSION.md:8`
- [ ] Document that staging uses the **same mainnet contracts, separate URL** (the prod/staging registry split was already removed — see `apps/web/src/lib/maps/contracts.ts` + `docs/MINIPAY_BUILD_PLAYBOOK.md`)
- [ ] Prune the leftover in-flux staging branches once the canonical flow lands (`remove-staging-flow`, `read-chain-sepolia-for-staging`, `minipay-learnings-and-staging`)

[...]

PR #212, opened 2026-08-10
fix(deps): repair the pnpm lockfile broken by two overlapping merges

GigaHierz:
## Urgent — `main` is red and Vercel cannot deploy

#207 and #203 both changed **only** `pnpm-lock.yaml` and were squash-merged back to back with no rebase between them. GitHub combined the two lockfiles textually and produced invalid YAML.

**CI:**
```
ERR_PNPM_BROKEN_LOCKFILE  The lockfile at "pnpm-lock.yaml" is broken:
duplicated mapping key (10796:3)
```

`/nanoid@3.3.18` and `/postcss@8.5.26` were each defined twice.

**Vercel** is the same cause wearing a different hat — unable to parse the lockfile, it couldn't detect the package manager, fell back to pnpm 6.35.1 and died on the engines check:
```
Skipping build cache since Package Manager changed from "pnpm" to "npm"
Error while parsing config file: "/vercel/path0/pnpm-lock.yaml"
ERR_PNPM_UNSUPPORTED_ENGINE  Expected version: >=8.0.0  Got: 6.35.1
```
That's not a separate pnpm-version problem — fix the lockfile and Vercel reads `packageManager: pnpm@8.10.0` again.

## The fix

I did **not** hand-patch the duplicates. There were several, and which copy is authoritative — the one with `dev: true` or without — is a guess I'd rather not encode.

Instead: restore `pnpm-lock.yaml` from **288eda7**, the last commit where it was valid, and reapply the postcss update with pnpm so the file is machine-generated and internally consistent.

**postcss lands on 8.5.26**, not the 8.5.23 in #203's title. Both satisfy the unchanged `^8.5.21` range and both carry the fix for CVE-2026-69153; 8.5.26 is what pnpm resolves to today. `package.json` was never touched by #203, so nothing else needed to change.

## Blast radius

The whole diff against the last-good lockfile:

```
+ /postcss@8.5.26  (and autoprefixer, postcss-import, postcss-js,
                    postcss-load-config, postcss-nested re-peered onto it)
- /postcss@8.5.21  (and the same five on the old peer)
```

**No runtime dependency moved.** I checked `viem`, `next`, `wagmi`, `@privy-io` and `react` explicitly, because regenerating a lockfile can silently re-resolve every caret range — that's how #89 quietly carries viem 2.47.4 → 2.55.11. Nothing here did.

## Verification

- Lockfile parses; `grep` for duplicate package keys returns nothing.
- **`pnpm install --frozen-lockfile` succeeds** — the exact command both CI and Vercel run, and the one that was failing.
- `type-check` clean.
- **425 tests pass.**
- **Production build green** — the real test, since postcss is build-time and CI doesn't build.

## Please read before merging the next lockfile PR

**#206 is still open and is also lockfile-only.** Merging it without rebasing onto this first reproduces this outage exactly. Same for #211 (coverage), which also touches the lockfile.

The general rule, until branch protection with "require branches to be up to date" lands in #199: **merge lockfile-only PRs one at a time, and let each one rebase before the next.** GitHub reports them all as `MERGEABLE` because it diffs them against `main` independently — it does not know two lockfiles are semantically incompatible.

### DIFF
--- a/apps/web/package.json
+++ b/apps/web/package.json
@@ -49,7 +49,7 @@
     "eslint": "^8.50.0",
     "eslint-config-next": "^14.0.0",
     "jsdom": "^29.0.0",
-    "postcss": "^8.5.21",
+    "postcss": "^8.5.26",
     "tailwindcss": "^3.4.4",
     "tailwindcss-animate": "^1.0.7",
     "typescript": "^5.2.2",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -122,7 +122,7 @@ importers:
         version: 6.0.3(vite@8.2.1)
       autoprefixer:
         specifier: ^10.4.17
-        version: 10.4.27(postcss@8.5.23)
+        version: 10.4.27(postcss@8.5.26)
       eslint:
         specifier: ^8.50.0
         version: 8.57.1
@@ -133,8 +133,8 @@ importers:
         specifier: ^29.0.0
         version: 29.0.0
       postcss:
-        specifier: ^8.5.21
-        version: 8.5.23
+        specifier: ^8.5.26
+        version: 8.5.26
       tailwindcss:
         specifier: ^3.4.4
         version: 3.4.19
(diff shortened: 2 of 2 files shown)


GigaHierz:
Correction to the description above: the PR **does** touch `apps/web/package.json`, one line, and it is required.

pnpm records the declared specifier next to the resolved version, so a lockfile regenerated at `^8.5.26` against a `package.json` still declaring `^8.5.21` fails the very check this PR exists to fix:

```
ERR_PNPM_OUTDATED_LOCKFILE  specifiers in the lockfile don't match specs in package.json
```

Both ranges accept 8.5.26 — only the recorded specifier differs. I caught this by trying to drop the change to keep the diff minimal, then testing rather than assuming: reverting that one line alone makes `--frozen-lockfile` fail, and restoring it makes it pass. Dropping it would have shipped a still-broken `main`.

Blast radius is otherwise unchanged: postcss and its five peers, no runtime dependency moved.

Issue #216, opened 2026-08-13
fix(minipay): all recent buy failures are MiniPay users — audit with per-category diagnostics


GigaHierz:
## Problem

PostHog shows every buy failure since Aug 8 comes from a MiniPay user, and July 23 recorded 113 MiniPay failures (67% of that day's total) against zero on July 22.

## Read the cliff correctly before auditing

The July 22 → 23 jump is almost certainly **when attribution started, not when a regression shipped**: #171 (merged the morning of July 23) is what added the `isMiniPay` super-property to PostHog. Before that, MiniPay failures existed but could not be segmented — so don't burn time diffing July 22 vs 23 looking for a phantom regression. Similarly, "100% of failures are MiniPay" needs a denominator: MiniPay is the primary surface, so compare MiniPay's failure share against its share of buy *attempts* before concluding it over-indexes.

Also note: #173 (same day) removed the provider-level MiniPay tx diagnostic that #151 had added, so today there are **no** MiniPay-specific diagnostics anywhere in the failure path.

## What to do

1. **After #215 lands** (failure `category` + raw `detail` on `pixel_buy_failed`), segment failures by `category` × `isMiniPay` — that turns "MiniPay fails a lot" into "MiniPay fails on X".
2. Compare MiniPay failure rate vs MiniPay share of `pixel_buy_started` to establish whether it actually over-indexes.
3. Audit the July 23 deploys that touched the buy path anyway (#171, #173, #174, #177) for MiniPay-specific behaviour changes — cheap to rule out.
4. Known MiniPay hazards to check first once categories exist: the CIP-64 fee-currency + explicit-gas path in `useBuyPixels.ts` (gas-estimate with `feeCurrency`, retry without, hard ceilings) and the adapter mapping in `src/lib/feeCurrency.ts`. A gas-less send makes MiniPay's `eth_estimateGas` return `permission denied` — currently invisible as a category.

## Depends on / related

- Depends on #215 (failure categories).
- Related MiniPay-funnel work: #170 (MiniPay sheet shows no spend amount).

GigaHierz:
## Steps 3 and 4 done — and hazard 4 is refuted

Working the code half while the data half waits for volume. Recording the negative
results precisely, because one of them rules out this issue's own prime suspect and I'd
rather nobody re-investigates it.

### The `feeCurrency` adapter mapping is correct — verified on-chain, not by reading

[...]

Issue #215, opened 2026-08-13
fix(analytics): pixel_buy_failed hides the root cause — 50.7% of failures are one generic bucket


GigaHierz:
## Problem

PostHog failure data shows **50.7% of all buy failures** land in one generic bucket: `pixel_buy_failed` with `reason: "That didn't go through — please try again."`. Half the failure volume is indistinguishable — chain reverts, gas problems, timeouts and RPC errors all look identical, so we can't see what to fix.

## Root cause

`track('pixel_buy_failed', { ...eventProps, reason: short })` in `apps/web/src/hooks/useBuyPixels.ts` sends the **user-facing string** as the only discriminator. The raw detail we already extract — `extractErrorDetail()` and the revert reason recovered via `simulateContract` — goes to `console.error` only and never reaches PostHog. `classifyBuyError` (`apps/web/src/lib/buyErrors.ts`) matches just six substrings (nonce, `NotLand`, `TokenNotAccepted`, `SlippageExceeded`, `DeadlineExpired`, insufficient/ERC20); everything else — including MiniPay's `permission denied` gas-estimation failure — falls through to the generic message.

Several failure modes emit **no event at all**:
- gas-estimate fallbacks (`console.warn` sites in `useBuyPixels.ts` around the approve/buy estimate blocks)
- pre-wallet guard errors: chain-switch rejected, no stablecoin balance, over-cap (`useBuyPixels.ts`, before the wallet prompt)

## What to change

1. Add a machine-readable `category` property to `pixel_buy_failed` alongside `reason` — e.g. `chain_revert`, `gas_estimate`, `permission_denied`, `timeout`, `rpc`, `nonce`, `insufficient_funds`, `not_land`, `slippage`, `unknown`.
2. Add a truncated raw `detail` field (the pattern already exists: `profile_save_failed` sends `msg.slice(0, 100)` in `useProfile.ts`).
3. Extend `classifyBuyError` with gas / timeout / MiniPay `permission denied` branches so those stop falling into `unknown`.
4. Emit events for the currently-invisible failures: gas-estimate fallback and the pre-wallet guards.
5. Update the event-schema docblock in `src/lib/analytics.ts` (it's already missing `pixel_buy_over_cap`, `profile_save_failed`, `activity_feed_shown`).

## Non-goals

[...]

Issue #193, opened 2026-08-04
Better buy/error handling (revert messages, retry, Blockscout links)


GigaHierz:
Improve trust on failed buys:
- map specific reverts to clear messages (e.g. NotLand, over-cap)
- offer a retry affordance
- link the failed tx to Blockscout

Today failures surface as generic errors.

csacanam:
Went through the current code against the three bullets here. Most of this has already shipped since the issue was written, so recording what's left before someone re-does the walk.

**Revert → clear messages: done.** `lib/buyErrors.ts` (from #161 and #173) maps `NotLand`, `TokenNotAccepted`, `SlippageExceeded`, `DeadlineExpired`, insufficient funds and nonce errors to short user-facing lines. Everything else — transient RPC blips, wallet quirks, raw viem dumps — collapses to one friendly retry line instead of a technical dump, with the raw error kept in the console.

**"Over-cap": done, but not as a revert.** There is no `OverCap` error in the contract — the $10 ceiling is client-side in `lib/buyLimits.ts` (#181) and disables the BUY button before anything is signed, so it never becomes a failed transaction. Nothing to map.

**Retry affordance: already there.** `SelectionDrawer` renders the full buy view for `txStep === 'idle' || 'error'`, so the BUY button comes back with the selection intact, and the fallback copy is literally *"That didn't work. Try again?"*.

**Blockscout link: not done.** This is the only bullet with real work left.

Two things worth knowing before picking it up:

1. **viem points Celo at celoscan.io**, not Blockscout — `celo.blockExplorers.default.url` is `https://celoscan.io`. So a Blockscout link can't be read off the chain config; it has to be declared separately.
2. **We only sometimes have a hash.** `useBuyPixels` sets `txHash` after the send (`useBuyPixels.ts:313`), so a transaction that reverts *on-chain* is linkable, but a failure at simulation, approval or in the wallet never produces one. The link has to be conditional on `txHash` or it will render broken for a large share of failures.

So the remaining scope is roughly: an explorer-URL helper with tests, plus a conditional link under the error line in `SelectionDrawer`.

One thing I would deliberately **not** do: map the remaining contract errors (`OutOfBounds`, `InvalidPixelId`, `NoTokens`, `InvalidToken`). They aren't reachable through the UI — you can't select zero pixels or an out-of-range id — so they'd be copy for states no player can reach.

GigaHierz:
[...]

Issue #170, opened 2026-07-22
Buy conversion: surface the spend amount before the wallet prompt (MiniPay sheet shows none)


GigaHierz:
## Context

First live MiniPay funnel shows two buy drop-offs: ~62% open the review drawer but never confirm a buy, and ~33% of those who start never reach a confirmed purchase (a chunk of which are silent wallet rejections). Surfaced during testing: the **MiniPay confirmation sheet shows no spend amount** — it reads "Buy pixels / Buy 3 pixels / Requested by Mondeto / Network fee" but never the dollars/USDT you're about to pay. Being asked to hit **Buy** with no visible price is a plausible driver of both the hesitation and the silent rejections.

### Why the amount is missing (confirmed from the code)

- The buy is a call to `buyPixels(ids, token, maxTotalCost, deadline)`. The USDT never moves in the calldata — the contract pulls it via the standing **allowance** (`transferFrom`) after the earlier `approve`. So the transaction MiniPay signs carries **no transfer amount**, only an internal `maxTotalCost` slippage bound in the contract's price units, which MiniPay does not surface.
- All the sheet text ("Buy pixels", "Tether USD", "Requested by Mondeto") is **MiniPay decoding the function name + token + origin** — none of it is passed by us.
- `wagmi`/`viem` `writeContractAsync` has no description/metadata channel, and we use only the injected connector (no MiniPay SDK). There is no field we can populate today to make the amount appear.

Relevant code: `src/hooks/useBuyPixels.ts` (`buyPixels` write ~L390-400, `approve` ~L329-343), `src/lib/contract.ts` (`buyPixels` ABI ~L156-181), `src/lib/feeCurrency.ts`.

## Scope: our own drawer, not the wallet sheet

We fully own the screen right before the wallet prompt. Ensure the precise amount + token is prominent at the moment of the CTA so nobody reaches MiniPay's blank sheet unsure what they'll pay.

The drawer already shows "THE DAMAGE" and "LOCK IT IN — {price} {token}" (`src/components/Overlays/SelectionDrawer.tsx`); tighten so the number is unambiguous immediately before the prompt, and consider a one-line "you'll pay {X} {token}" restated right at the CTA, plus a two-prompt (approve + buy) explainer for first-time buyers.

## Explicitly not in this issue

[...]

PR #220, opened 2026-08-14
fix(analytics): give buy failures a stable category and emit the ones that were silent

csacanam:
## Problem

**50.7% of all buy failures land in one bucket** — `pixel_buy_failed` with
`reason: "That didn't go through — please try again."` Half the failure volume is
indistinguishable: chain reverts, gas problems, timeouts and RPC blips all look
identical, so there is nothing to act on.

Verified against the code before changing anything:

- `useBuyPixels.ts` sent the **player-facing string** as the only discriminator.
- The raw detail we already extract (`extractErrorDetail()`, and the revert reason
  recovered via `simulateContract`) went to `console.error` only.
- `classifyBuyError` matched exactly six substrings. MiniPay's `permission denied`
  — the most likely cause of the MiniPay failure concentration in #216 — was not
  among them, so it fell into the generic bucket.
- **Seven paths emitted no event at all**: four gas-estimate fallback rungs and
  three pre-wallet guards.

## What changes

**`lib/buyErrors.ts`** — the six substring checks become an ordered rule table.
`classifyBuy()` returns the message and the category in a single pass, so the pair
can't drift. The original six rules keep their exact matching and copy; five new
categories (`permission_denied`, `timeout`, `rpc`, `gas_estimate`, `chain_revert`)
split what used to collapse into `unknown` — all of them keeping
`GENERIC_RETRY_MESSAGE`, so **no player-facing copy changes in this PR**.

Rule order is load-bearing and tested: viem wraps a transient provider blip as
`... reverted with the following reason: RPC endpoint returned HTTP client error`,
so `rpc` is tested before `chain_revert` — otherwise every Forno rate-limit would
be filed as an on-chain revert and send someone to debug the contract.

**`hooks/useBuyPixels.ts`** — `pixel_buy_failed` gains `category` and a `detail`
truncated to 100 chars (the pattern `profile_save_failed` already uses). The seven
silent paths now emit.

**`lib/analytics.ts`** — the event-schema docblock is brought back in sync: it was
missing `pixel_buy_over_cap`, `profile_save_failed` and `activity_feed_shown`, and
still listed `invite_shared`, which no longer exists (superseded by
`share_clicked { kind: 'invite' }`).

## Two calls worth a look at review

**The pre-wallet guards get their own event, not `pixel_buy_failed`.** All three
fire *before* `track('pixel_buy_started')`, so filing them as failures would produce
failures with no corresponding start and break that funnel's arithmetic. They emit
`pixel_buy_blocked` instead — never preceded by a start, never followed by a
failure, safe to sum separately. This is documented in the docblock, including the
warning not to add them to a failure rate whose denominator is `pixel_buy_started`.

**Gas fallbacks are not failures.** The buy usually still goes out, so
`pixel_buy_gas_fallback` carries `stage` (approve/buy) and `level`
(without_fee_currency/ceiling) rather than being counted as a loss. That rung is the
tell for the MiniPay CIP-64 hazard #216 names.

## Worth knowing for #170

The three pre-wallet guards were not merely missing a failure event — because they
fire before `pixel_buy_started`, and `handleBuy` (`app/page.tsx:359`) calls
`execute()` directly, a player who opens the drawer, taps BUY and hits a guard
produces a `checkout_opened` and then nothing. They are indistinguishable from
someone who changed their mind.

`SelectionDrawer.tsx:106-109` says the pre-buy drop-off is instrumented so it "can
be split into named causes instead of one opaque '62% left'". These three guards
were the unnamed remainder of exactly that 62% — #170's headline number. From here
they emit `pixel_buy_blocked` with a reason.

To be explicit about what this is *not*: it does **not** bias the failure rate #216
plans to use. A blocked buy is absent from both sides of
`pixel_buy_failed / pixel_buy_started`, so that ratio stays internally consistent
and #216's step 2 comparison is sound as written. What was hidden is a separate
population, which is a conversion question rather than a failure-rate one.

## Why it's safe

The six original classifications are unchanged in both matching and copy — the three
pre-existing tests pass untouched. `eventProps` is now assembled in two steps but
resolves to the same five fields, so `pixel_buy_started`, `pixel_buy_approve_shown`,
`pixel_buy_succeeded`, `pixel_buy_rejected` and `pixel_buy_over_cap` keep their exact
prop shape and no existing funnel metric moves. No transaction-building code was
touched; the only behavioural change is which events are emitted.

## Verification

- `tsc --noEmit` clean.
- **432 tests pass** across 52 files (13 in `buyErrors.test.ts` — the 3 original plus
  10 new).
- Emitted-vs-documented event diff: **0 emitted without documentation, 0 documented
  without emission.**

New tests cover the rule ordering specifically: that an RPC blip categorises as `rpc`
and not `chain_revert`, that `insufficient funds for gas` stays `insufficient_funds`,
that the new categories match case-insensitively (MiniPay's wording is lowercase,
viem's is not), and that the message and category can never disagree.

**Not verified automatically:** the event emission in `useBuyPixels.ts`. The hook has
no test coverage by design — `buyErrors.ts` was extracted precisely so classification
could be tested without mounting wagmi/React. The three `pixel_buy_blocked` paths are
free to trigger on the preview (select over \$10; connect a wallet with no stablecoin;
reject the chain switch) and each sits immediately above a `setError` that already
works today, with no branch between them. `pixel_buy_gas_fallback` and a real
`category` need a genuine failure and are left unverified — the first query in #216
confirms them at no extra cost.

## Non-goals

User-facing error copy (retry, revert messages, Blockscout links) stays in #193 — the
`category` taxonomy built here is what that work should consume. No player-visible
string changes in this PR.

Refs #215

### DIFF
--- a/apps/web/src/__tests__/hooks/useBuyPixels.test.ts
+++ b/apps/web/src/__tests__/hooks/useBuyPixels.test.ts
@@ -2,6 +2,7 @@ import { describe, it, expect, vi } from 'vitest'
 import { renderHook, act } from '@testing-library/react'
 import { useBuyPixels } from '@/hooks/useBuyPixels'
 import { OVER_SPEND_CAP_MESSAGE, PRICE_MOVED_MESSAGE } from '@/lib/buyLimits'
+import { track } from '@/lib/analytics'
 
 // Shared, stable write mock so the cap tests can assert the wallet is never
 // opened. Hoisted above the vi.mock factory (which is itself hoisted). The
@@ -67,7 +68,9 @@ vi.mock('@/hooks/useStablecoinBalance', () => ({
 }))
 
 // Keep analytics inert — the execute() path fires funnel events we don't want
-// hitting PostHog in unit tests.
+// hitting PostHog in unit tests. `track` is imported below so the emissions
+// themselves can be asserted: the pre-wallet guards are the paths that were
+// invisible, so "it blocked" is only half of what needs proving.
 vi.mock('@/lib/analytics', () => ({ track: vi.fn(), getReferrer: () => undefined }))
 
 describe('useBuyPixels', () => {
@@ -101,6 +104,18 @@ describe('useBuyPixels', () => {
     expect(result.current.error).toBe(OVER_SPEND_CAP_MESSAGE)
     // Never reached the approve/buy writes.
     expect(writeContractAsync).not.toHaveBeenCalled()
+
+    // The guard has to be *counted*, not just enforced. Blocks like this fire
+    // before `pixel_buy_started`, so they were invisible entirely — not merely
+    // missing a failure event, but absent from the attempt denominator too.
+    expect(track).toHaveBeenCalledWith(
+      'pixel_buy_blocked',
+      expect.objectContaining({ reason: 'over_spend_cap' }),
+    )
+    // And it must not double as a failure: `pixel_buy_failed` with no matching
+    // `pixel_buy_started` would corrupt the very funnel this work exists to fix.
+    expect(track).not.toHaveBeenCalledWith('pixel_buy_failed', expect.anything())
+    expect(track).not.toHaveBeenCalledWith('pixel_buy_started', expect.anything())
   })
 
   it('blocks with a "prices moved" nudge when the live price tips a sub-$10 pick over the cap', async () => {
--- a/apps/web/src/__tests__/lib/buyErrors.test.ts
+++ b/apps/web/src/__tests__/lib/buyErrors.test.ts
@@ -98,6 +98,69 @@ describe('categorizeBuyError', () => {
     expect(categorizeBuyError('Execution Reverted')).toBe('chain_revert')
   })
 
+  it('files a real revert as chain_revert even inside a full viem envelope', () => {
+    // The fixture that matters. viem appends `Docs: https://viem.sh…` to any
+    // error with a docsPath — writeContract, simulateContract and
+    // estimateContractGas all set one — so a bare `http` match would swallow
+    // this and file a genuine on-chain revert as a transport blip. Bare
+    // fragments like 'execution reverted' pass either way and prove nothing.
+    const hay = [
+      'The contract function "buyPixels" reverted.',
+      '',
+      'Error: NotEnoughSomething()',
+      '  Contract Call:',
+      '    address:   0x1234567890123456789012345678901234567890',
+      '    function:  buyPixels(uint256[] ids, address token, uint256 maxTotalCost, uint256 deadline)',
+      '    sender:    0x9876543210987654321098765432109876543210',
+      '',
+      'Docs: https://viem.sh/docs/contract/simulateContract',
+      'Version: viem@2.21.0',
+    ].join('\n')
+    expect(categorizeBuyError(hay)).toBe('chain_revert')
+  })
+
+  it('files a genuine transport failure as rpc, envelope and all', () => {
+    const hay = [
+      'HTTP request failed.',
+      '',
+      'Status: 429',
+      'URL: https://lb.drpc.org/ogrpc?network=celo',
+      'Request body: {"method":"eth_call"}',
+      '',
+      'Details: Too Many Requests',
+      'Version: viem@2.21.0',
+    ].join('\n')
+    expect(categorizeBuyError(hay)).toBe('rpc')
+  })
+
+  it('does not read the fallback RPC hostname as an RPC failure', () => {
+    // `lb.drpc.org` carries the substring `rpc`, so a bare token match files
+    // every error routed through the fallback endpoint as a transport problem.
+    // Deliberately a revert with NO named error in the list above — otherwise
+    // an earlier rule claims it and the fixture proves nothing either way.
+    const hay = [
+      'The contract function "buyPixels" reverted.',
+      'Error: CustomErrorNobodyMapped()',
+      'URL: https://lb.drpc.org/ogrpc?network=celo',
+      'Version: viem@2.21.0',
+    ].join('\n')
+    expect(categorizeBuyError(hay)).toBe('chain_revert')
+  })
+
+  it('does not read a gas field in a revert envelope as an estimation failure', () => {
+    // Again no named error: `gas: 300000` in the request arguments is what a
+    // bare `gas` match would seize on, and `gas_estimate` is ordered ahead of
+    // `chain_revert`.
+    const hay = [
+      'The contract function "buyPixels" reverted.',
+      'Error: CustomErrorNobodyMapped()',
+      '  Request Arguments:',
+      '    gas:  300000',
+      'Docs: https://viem.sh/docs/contract/writeContract',
+    ].join('\n')
+    expect(categorizeBuyError(hay)).toBe('chain_revert')
+  })
+
   it('files a transient RPC blip as rpc, not as an on-chain revert', () => {
     // viem wraps provider hiccups as a "reverted" message. If chain_revert were
     // tested first, every Forno rate-limit would be miscounted as a contract
--- a/apps/web/src/__tests__/lib/buyErrors.test.ts
+++ b/apps/web/src/__tests__/lib/buyErrors.test.ts
@@ -1,5 +1,11 @@
 import { describe, it, expect } from 'vitest'
-import { classifyBuyError, isUserRejectedError, GENERIC_RETRY_MESSAGE } from '@/lib/buyErrors'
+import {
+  categorizeBuyError,
+  classifyBuy,
+  classifyBuyError,
+  isUserRejectedError,
+  GENERIC_RETRY_MESSAGE,
+} from '@/lib/buyErrors'
 
 describe('isUserRejectedError', () => {
   it('detects EIP-1193 code 4001 (top-level and nested cause)', () => {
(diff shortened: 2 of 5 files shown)


GigaHierz (changes requested):
The structure here is right and the funnel reasoning is careful. One rule in the table
undoes the point of the change, though, and because these category names become a
permanent PostHog dimension it's worth fixing before it starts writing data.

### Strengths

- **The structural change is the right one.** Collapsing six scattered `if`s into an
  ordered `RULES` table with a single `classifyBuy()` pass, and deriving
  `classifyBuyError`/`categorizeBuyError` as views of it, makes it structurally impossible
  for the player-facing message and the analytics category to drift apart. The
  `never disagrees with the message it was classified alongside` test locks that in.
- **Requirement coverage is complete on all five asks**, and (5) is not just updated but
  *correct* — every `track('…')` call site in `src/**` was diffed against the docblock and
  the "0 emitted without documentation, 0 documented without emission" claim holds across
  all 28 events.
- **Splitting pre-wallet guards into `pixel_buy_blocked` rather than `pixel_buy_failed` is
  the right call**, and the docblock's reasoning is correct: they fire before
  `pixel_buy_started`, so filing them as failures would corrupt that funnel's denominator
  — and the 50.7% number this work exists to fix.
- **No double-counting.** Traced `execute()`: exactly one `track('pixel_buy_failed')` call
  site, in the single `catch`; `pixel_buy_over_cap` and all three guards `return` before
  reaching it; `inFlight` prevents a re-entrant pass. One user-visible failure produces
  exactly one failure event, and the gas-fallback events use a separate name so they can't
  inflate the ratio.
- **No user-facing copy regression** — all six original messages are byte-identical in the
  rule table and `classifyBuyError`'s signature is preserved, so #193's scope is respected.

### Issues

#### Critical

**1. The `rpc` rule matches the bare substring `http`, which appears in essentially every
viem error message — so it swallows the reverts it was ordered ahead of.**

`lib/buyErrors.ts` —

```ts
category: 'rpc',
test: (_hay, lower) =>
  lower.includes('rpc') ||
  lower.includes('http') ||          // ← matches almost everything
  ...
```

viem's `BaseError` appends `Docs: https://viem.sh<docsPath>` to `message` whenever
`docsPath` is set, and `writeContract`, `simulateContract` and `estimateContractGas` all
set it. `RpcRequestError` also prepends `URL: https://<rpc>`. Since `hay = e.message + ' '
+ detail`, `lower.includes('http')` is true for practically every viem-originated error
this hook can catch.

**Confirmed against a real production error rather than a reconstruction.** An operator
hit this exact shape on the admin payout screen today:

```
The contract function "disperseToken" reverted with the following reason:
ERC20: transfer amount exceeds allowance
...
Docs: https://viem.sh/docs/contract/simulateContract
Details: execution reverted: ERC20: transfer amount exceeds allowance
Version: viem@2.53.1
```

That is an unambiguous on-chain revert, and it contains `https://viem.sh`. Run it through
this table and it classifies as **`rpc`**, not `chain_revert` — because `rpc` sits at
position 9 and `chain_revert` at position 11.

The consequence is the inverse of what the PR body argues. The body says `rpc` is tested
before `chain_revert` so a Forno blip isn't misfiled as a contract revert; in practice
**real contract reverts are misfiled as `rpc`**, and `gas_estimate` and `chain_revert`
become near-dead branches — the only things reaching `chain_revert` are the two strings
the hook throws itself at `useBuyPixels.ts:371,377`, and the first survives only because
`.slice(0, 150)` happens to cut before the `Docs:` line.

So on merge, `rpc` immediately becomes the new ~50% generic bucket — the same problem
#215 was filed to fix, relabelled — and #216 would be built on top of it. Re-bucketing
later orphans the historical data, which is why this is worth blocking on.

*Fix:* drop the bare `http` and `rpc` substrings and match signals that actually indicate
transport failure:

```ts
test: (_hay, lower) =>
  lower.includes('http request failed') ||
  lower.includes('http client error') ||
  lower.includes('too many requests') ||
  lower.includes('rate limit') ||
  lower.includes('failed to fetch') ||
  lower.includes('fetch failed') ||
  lower.includes('network error') ||
  lower.includes('unknown rpc error'),
```

`lower.includes('gas')` has the same shape of problem (viem's `Request body` meta-message
contains a `gas` key) and should narrow to `intrinsic gas` / `out of gas` /
`gas required exceeds` / `gas estimate`. "insufficient funds for gas" is already claimed
by the rule above it, so that narrowing is safe.

If you'd rather not play substring whack-a-mole: strip viem's envelope (`Docs:`,
`Version:`, `URL:`, `Request body:`, the `Contract Call:` block) out of `hay` before
matching, or classify against `shortMessage` + `details` rather than the fully formatted
`message`. That's the more durable answer.

**2. The tests assert against synthetic strings, which is why (1) got through.**

`categorizeBuyError('execution reverted')` and `categorizeBuyError('intrinsic gas too
low')` pass because they're bare fragments with no viem envelope — a shape that never
occurs in production. The `rpc`-before-`chain_revert` test passes for the same wrong
reason.

*Fix:* add one fixture that is a full viem message (short message + `Contract Call:` block
+ `Docs:` + `Version:`) asserting `chain_revert`, and one full `HTTP request failed` /
`Status: 429` envelope asserting `rpc`. Those two fixtures are what make the taxonomy
trustworthy going forward. The real string quoted above is a ready-made fixture.

#### Important

**3. `trackGasFallback` sends the wrapped error, not the unwrapped one — so the event most
likely to explain the MiniPay CIP-64 failure carries boilerplate instead of the reason.**
`useBuyPixels.ts:158-165` uses raw `.message`. `extractErrorDetail()` exists in this very
file precisely because "viem masks provider failures as *An unknown RPC error occurred* at
the top level while the real reason lives in `.cause`/`.details`/`.data`". For a MiniPay
estimate failure the first 100 chars are roughly `An unknown RPC error occurred.\n\nURL:
https://<forno>\nRequest body: {"method":"eth_estim…` — the `permission denied` string
that makes this event worth having sits in `Details:`, well past the cut.
`pixel_buy_failed` gets this right; this call site should too:
`detail: extractErrorDetail(err).slice(0, 100)`.

Secondary effect: because `URL: <rpc>` lands inside the first 100 chars, the authenticated
`NEXT_PUBLIC_FORNO_RPC_URL` gets written into PostHog properties. Not a new secret — it's
already inlined in the client bundle and protected by domain allowlist rather than
secrecy — but there's no reason to propagate it into a third-party store, and switching to
`extractErrorDetail` removes it as a side effect.

**4. The body's reason for not testing the emissions is factually wrong.** It says *"The
hook has no test coverage by design."* `src/__tests__/hooks/useBuyPixels.test.ts` exists
(129 lines), mocks `@/lib/analytics` with `track: vi.fn()`, and already drives `execute()`
through the very guard this PR instruments — `blocks a purchase over the $10 cap before
opening the wallet`. Asserting `track` was called with `'pixel_buy_blocked',
expect.objectContaining({ reason: 'over_spend_cap' })` is a two-line addition to an
existing test; `no_stablecoin_balance` needs only a mock flip. Please add those or correct
the claim.

#### Minor

**5. `pixel_buy_gas_fallback` can fire twice per stage, up to four times per attempt.**
`level: 'ceiling'` is always preceded by `level: 'without_fee_currency'` for the same
stage, because the ceiling branch is nested inside the retry's `catch`. A dashboard
counting raw events overstates affected buys ~2×. Worth a sentence in the docblock: count
`without_fee_currency`; `ceiling` is a strict subset.

**6. `chain_switch_rejected` fires on any `switchChainAsync` throw, not only user
rejection** (`useBuyPixels.ts:124` — bare `catch`). A wallet failing
`wallet_addEthereumChain` outright is filed as a rejection. `chain_switch_failed` is the
honest name, or split on `isUserRejectedError`, already imported.

### Assessment

**Ready to merge?** No

**Reasoning:** The plumbing, event design and funnel arithmetic are all right, but the
`rpc` rule's bare `http` match means real on-chain reverts and gas failures get filed as
`rpc` — so the new taxonomy relabels the generic bucket instead of splitting it, and these
names are permanent once they start landing in PostHog. Issues 1–3 are perhaps 20 lines,
not a redesign; with those plus one realistic-envelope test this is a clear approve.

csacanam:
All six addressed. The critical one was worse than the review found — the same trap appears twice more in the same table.

## 1 · The `http` match — and two siblings

Fixed, and while fixing it I checked the other two rules for the same shape:

- **`lower.includes('rpc')`** has the identical problem from the other direction. The fallback endpoints from #113 include `lb.drpc.org`, so the substring rides along in the `URL:` line of any error routed through them — filing every one as a transport failure.
- **`lower.includes('gas')`** likewise: viem prints `Request Arguments:` blocks carrying `gas:`, so the bare token appears in errors that have nothing to do with estimation.

All three now match multi-word phrases that cannot occur in a URL — `http request failed`, `rpc endpoint`, `status: 429`, `estimate gas`, `intrinsic gas`, and so on.

## 2 · The fixtures, and a step I owe you

Replaced with full viem envelopes, and then did the thing I should have done first: **stashed the fix and ran the new tests against the old rules.** Three fail. That's what makes them worth having.

It also caught that my first two replacements were still worthless — they carried `NotLand` and `SlippageExceeded`, which earlier rules claim, so they passed either way. Exactly the failure you'd named, reproduced one layer up. Both now use an unmapped custom error so nothing else can claim them.

Before / after, same fixtures:

```
new tests + old rules  →  3 failed, 14 passed
new tests + new rules  →  17 passed
```

## 3 · `trackGasFallback` now sends `extractErrorDetail(err)`

You were right that the event most likely to explain the MiniPay concentration was carrying boilerplate. The RPC URL dropping out of PostHog is a welcome side effect.

## 4 · The claim was wrong, and the assertions are in

`useBuyPixels.test.ts` does exist and already drove the cap guard — I asserted the opposite in the body without checking. The over-cap test now asserts `pixel_buy_blocked` with `reason: 'over_spend_cap'`, **and** that neither `pixel_buy_failed` nor `pixel_buy_started` fires, since a blocked buy appearing in either would corrupt the funnel this work exists to fix.

**Heads-up on a collision:** #236 rewrites `useBuyPixels.test.ts` from 76 to 220 lines and doesn't cover these emissions, so whichever of us merges second will need to reapply this. Flagging rather than discovering it later.

## 5 and 6 · Both done

`ceiling` is documented as a strict subset of `without_fee_currency` in the event schema, with the ~2x warning. `chain_switch_failed` splits from `chain_switch_rejected` via `isUserRejectedError`, which was already imported — a wallet that can't add Celo is actionable in a way a decline isn't.

## Verification

`tsc --noEmit` clean, **449 tests pass** across 55 files. Rebased onto `main` (it was BEHIND after #229).

No player-facing copy moved: all six original messages remain byte-identical and the five added categories all keep `GENERIC_RETRY_MESSAGE`, so #193's scope is still untouched.

GigaHierz (approved):

Issue #221, opened 2026-08-15
mondeto.app crashes on load in every browser except MiniPay — down since Aug 10


csacanam:
## Summary

`mondeto.app` renders **"Application error: a client-side exception has occurred"** instead of the app. Production has been serving the affected build since **2026-08-10 21:24 UTC — four days**.

Confirmed on `/` and `/ranks`. Both fail identically.

## Evidence

Reproduced in a clean browser with no extensions and no session, and independently in a normal browser session, so it is not local to one machine:

```
`useWallets` was called outside the PrivyProvider component     (warning, ×2)
TypeError: Cannot read properties of undefined (reading 'current')   (uncaught)
    at a (chunks/5541.fb90f59232b1c026.js:62:142397)
    at es (…:257:8033)
    …
```

The uncaught exception is what takes the tree down; the two warnings precede it in the same chunk.

## Why MiniPay is very likely unaffected

`wallet-provider.tsx:91` branches on `isMiniPay` **alone** — there is no desktop/mobile distinction:

```ts
if (!mounted || isMiniPay) {
  // vanilla wagmi tree — no Privy module on this path
}
// everyone else: <PrivyTree>{children}</PrivyTree>
```

MiniPay clients keep the vanilla tree and never mount Privy, so they cannot reach the crashing code. Everyone else does.

This is an inference from the code, **not verified on a device** — MiniPay runs in the system WebView and I can't test it from here. Worth someone confirming on a real handset, because the whole severity assessment turns on it.

It is also the likely reason this went unnoticed for four days: we test in MiniPay.

## Prime suspect: #209

`bbc1801` — *perf(minipay): stop shipping Privy, x402 and @solana/kit to clients that never use them* — went to production at 21:24 on Aug 10, which matches the window exactly.

It is the **only** commit since July 3 touching `wallet-provider.tsx`, `wallet-provider-privy.tsx` or `connect-button.tsx`, and the crash sits precisely at the `PrivyProvider` boundary that PR restructured.

**Not confirmed.** I tried to bisect against a pre-#209 preview deployment and couldn't — preview URLs redirect to production. Someone with Vercel access to `minipay-celo-s-projects` can settle it in a minute by loading the deployment before `bbc1801`.

[...]

PR #222, opened 2026-08-15
chore(bisect): revert #209 to test whether it causes the browser crash (diagnostic, do not merge)

csacanam:
**Diagnostic only. Do not merge.** This exists to answer one question for #221 and will be closed either way.

## What this tests

#221: `mondeto.app` crashes on load in every browser (MiniPay verified fine on device). Prime suspect is #209 — the only commit since July 3 touching `wallet-provider.tsx`, `wallet-provider-privy.tsx` or `connect-button.tsx`, and it reached production at 21:24 on Aug 10, matching the window exactly.

This branch is `origin/main` with `bbc1801` (#209) reverted, nothing else. Load its Vercel preview:

- **Preview renders the map** → #209 is confirmed as the cause.
- **Preview still shows "Application error"** → the suspect is wrong, and the search widens to the other seven commits between `4e8fcde` and `fe764bd`.

## Why this must not be merged

Reverting #209 puts Privy, `x402` and `@solana/kit` back into MiniPay's bundle — the exact problem #209 was solving, and the one #196 traced to old WebViews failing to parse the chunk at all. Trading a browser outage for a MiniPay outage is not a fix.

Confirming the cause and choosing the remedy are separate decisions. This PR only does the first.

## Note on the revert

It applied cleanly against current `main`, so the six files return to their pre-#209 state: `connect-button.tsx` back to a single component, and `connect-button-interactive.tsx`, `connect-button-styles.ts`, `privy-ready-context.ts` plus `minipay-privy-isolation.test.ts` removed.

Refs #221


csacanam:
Answered: **#209 is not the cause.** This branch (main with #209 reverted) crashes identically — same `useWallets` warning, same `TypeError: Cannot read properties of undefined (reading 'current')`.

Result recorded in #221. Closing, since a revert of #209 is now ruled out as a remedy too: it wouldn't fix the outage and would put Privy + x402 + @solana/kit back into MiniPay's bundle.

Issue #182, opened 2026-07-29
/api/pnl overstates EARNED — never subtracts the 5% resale fee


GigaHierz:
## Problem

`apps/web/src/app/api/pnl/route.ts:84` credits the seller the buyer's **full** per-pixel price:

```ts
const perPixelCost = ids.length > 0 ? normalized / BigInt(ids.length) : 0n
for (const id of ids) {
  const idStr = id.toString()
  if (ownerOf.get(idStr) === addr) totalEarned += perPixelCost   // <- gross, no fee deducted
  ownerOf.set(idStr, buyer)
}
```

On-chain, a seller receives `price - fee`, where `feeRate` is **500 bps (5%)** — verified live with `cast call <proxy> "feeRate()(uint256)"` → `500` on the World, Africa and Europe maps. The route never applies it, so the EARNED tile on the profile page overstates what actually reached the wallet by ~5%.

A player who received $11.40 sees $12.00.

## Why it matters

This is its own support-ticket generator, in the same family as the payout-timing confusion just addressed in the FAQ rewrite: the app states a number that the player's wallet contradicts. Unlike those, this one is a genuine correctness bug rather than a copy problem, which is why it was kept out of that PR.

## Secondary issue

Line 81 splits a batch's `totalCost` evenly across all `ids` rather than pricing each pixel individually. Since per-pixel prices vary with each pixel's own `saleCount` and decay, per-pixel attribution is approximate for any multi-pixel buy. This skews both SPENT and EARNED whenever a batch mixes cheap and expensive pixels.

## Suggested fix

Deduct the fee when crediting a seller. `feeRate` is a storage variable (admin-settable, capped at 2000 bps), so read it rather than hardcoding — `apps/web/src/app/api/analytics/route.ts` already reads it from the same `config()` call and could share a helper.

Note `feeRate` applies only to **resales**; the full price of a previously unowned pixel goes to the treasury, and that path already contributes nothing to `totalEarned`, so only the seller branch needs changing.

csacanam:
Fix in #223. One correction to the issue while it's fresh: **the path named here isn't the one serving production.**

The body points at `api/pnl/route.ts:84`, the log-scan fallback. That code is real and does credit gross — but it only runs when the subgraph is unconfigured. The subgraph path has the same defect and is the live one:

`apps/subgraph/src/mapping.ts:180`

```ts
earnedBy.set(prevOwnerId, prevEarned.plus(perPixelCost))   // gross
```

[...]

PR #223, opened 2026-08-15
fix(profile): report EARNED net of the resale fee, not the price the buyer paid

csacanam:
## Problem

On chain a seller receives `price − fee`, where `feeRate` is **500 bps** — verified live on the World, Africa and Europe maps. Both P&L paths report the buyer's **gross** price instead, so the EARNED tile on the profile overstates what actually reached the wallet by the fee rate.

A player paid **$11.40** is shown **$12.00** — a number their own transaction history contradicts. Same family as the payout-timing confusion the FAQ rewrite addressed, except this one is a correctness bug rather than a copy problem.

## The issue understates it — both paths are affected, and the wrong one is named

#182 points at the log-scan fallback in `api/pnl/route.ts`. That path is real, but it only runs when the subgraph is unconfigured. **The subgraph path has the identical defect and is the one serving production:**

`apps/subgraph/src/mapping.ts:180`

```ts
earnedBy.set(prevOwnerId, prevEarned.plus(perPixelCost))   // gross
```

That accumulates into `Owner.totalEarned`, which `fetchOwnerPnl` returns verbatim. The comment immediately above the crediting loop says previous owners "earn the **fee-less** per-pixel proceeds" — the comment is wrong; no fee is ever applied. The mapping does track `MapStat.feeRateBps`, but only `/api/analytics` consumes it, for treasury revenue.

So fixing only the path the issue names would have left production untouched.

## What changes

**`lib/resaleFee.ts`** (new) — `readFeeRateBps`, lifted verbatim out of `/api/analytics` so the two routes share one definition rather than drifting, plus a pure `netOfResaleFee(gross, feeRateBps)`.

**`api/pnl/route.ts`** — nets EARNED after either source returns, so one correction covers both paths.

**`api/analytics/route.ts`** — imports the shared helper; its local copy is deleted. No behaviour change.

SPENT is deliberately untouched: the buyer really does pay the whole price.

## Why net in the route rather than in the subgraph

The subgraph is where the number is wrong at source, and fixing it there is the more correct end state. It is **not** what this PR does, for one reason: correcting the mapping only affects data indexed after a **full resync from the start block**. Every wallet's existing lifetime total would stay overstated until that completed, so the bug the player sees would persist through the fix.

Netting in the route corrects already-indexed history immediately, with no resync and no subgraph deploy. Worth opening the mapping fix as a follow-up so the two eventually agree — at which point this route-level netting has to come out in the same change, or the fee would be applied twice.

## Two approximations, both documented in the code

- **The current rate is applied to lifetime totals.** If `feeRate` has ever changed, older sales settled at the old one. `/api/analytics` already makes exactly this trade for treasury revenue; matching it keeps the two figures consistent with each other, which matters more here than either being individually exact.
- **Truncation.** On chain the fee is truncated per transfer; this applies it once to an aggregate. Bounded by a microcent per resale. `netOfResaleFee` takes the fee then subtracts, mirroring the contract's own shape, so the truncation at least lands the same direction.

## Why it's safe

A failed `feeRate` read returns 0 and reports gross — the pre-existing behaviour, and it understates the correction rather than over-deducting. A nonsensical rate (negative, ≥100%, `NaN`) is likewise treated as 0, so a bad read can never flip a player's earnings negative. Both are tested.

The arithmetic is `bigint` throughout; a float would already be lossy at real lifetime magnitudes.

`/api/analytics` keeps its exact behaviour — the helper moved, its body didn't.

## Verification

- `tsc --noEmit` clean.
- **432 tests pass** across 53 files, 7 of them new in `resaleFee.test.ts`: the $12.00 → $11.40 case from the issue, rate independence at 1000 and 2000 bps, truncation at the boundary, the bad-rate guards, zero, and a lifetime-scale total.

**Not verified automatically:** that the profile tile renders the corrected figure end to end. It needs a wallet with resale history, and the arithmetic is covered by unit tests. Worth one look on the preview by anyone who has sold a pixel.

## Non-goals

The subgraph mapping fix, and its wrong comment, are left for the follow-up described above rather than widened into here.

Refs #182

### DIFF
--- a/apps/web/src/__tests__/lib/resaleFee.test.ts
+++ b/apps/web/src/__tests__/lib/resaleFee.test.ts
@@ -1,0 +1,51 @@
+import { describe, it, expect } from 'vitest'
+import { netOfResaleFee } from '@/lib/resaleFee'
+
+// The live rate on the World, Africa and Europe maps at the time of writing,
+// verified with `cast call <proxy> "feeRate()(uint256)"`.
+const LIVE_FEE_BPS = 500
+
+describe('netOfResaleFee', () => {
+  it('takes the 5% cut the contract actually keeps', () => {
+    // The case from #182, in the units the route works in: a player who was
+    // credited $12.00 gross received $11.40.
+    expect(netOfResaleFee(12_000_000n, LIVE_FEE_BPS)).toBe(11_400_000n)
+  })
+
+  it('reports gross when there is no fee', () => {
+    expect(netOfResaleFee(12_000_000n, 0)).toBe(12_000_000n)
+  })
+
+  it('scales with the rate rather than assuming 5%', () => {
+    // feeRate is admin-settable up to 2000 bps, so nothing may hardcode 500.
+    expect(netOfResaleFee(1_000_000n, 1_000)).toBe(900_000n)
+    expect(netOfResaleFee(1_000_000n, 2_000)).toBe(800_000n)
+  })
+
+  it('truncates the fee the way the contract does', () => {
+    // Fee first, then subtract — the contract's own shape. Computing
+    // `gross * (10000 - bps) / 10000` instead would round the other way and
+    // reintroduce the discrepancy this exists to remove.
+    expect(netOfResaleFee(199n, LIVE_FEE_BPS)).toBe(190n) // fee 9.95 → 9
+    expect(netOfResaleFee(1n, LIVE_FEE_BPS)).toBe(1n) //     fee 0.0005 → 0
+  })
+
+  it('never invents a debt from a bad rate', () => {
+    // A failed read or a nonsensical value must understate the correction, not
+    // flip a player's earnings negative on their profile.
+    for (const bad of [-1, 10_000, 99_999, Number.NaN, Number.POSITIVE_INFINITY]) {
+      expect(netOfResaleFee(1_000_000n, bad)).toBe(1_000_000n)
+    }
+  })
+
+  it('leaves zero at zero', () => {
+    // A wallet that never sold anything shows 0, not a rounding artefact.
+    expect(netOfResaleFee(0n, LIVE_FEE_BPS)).toBe(0n)
+  })
+
+  it('holds at the scale of real lifetime totals', () => {
+    // No precision loss: these are bigints, and a float would already be
+    // wrong at this magnitude.
+    expect(netOfResaleFee(1_234_567_890_123n, LIVE_FEE_BPS)).toBe(1_172_839_495_617n)
+  })
+})
--- a/apps/web/src/app/api/analytics/route.ts
+++ b/apps/web/src/app/api/analytics/route.ts
@@ -4,6 +4,7 @@ import { MONDETO_ABI } from '@/lib/contract'
 import { getContractByMapId } from '@/lib/maps/contracts'
 import { estimateHistoryFromBlock, scanNormalizedPurchases, toMicrocents } from '@/lib/purchaseLogs'
 import { fetchBatchesSince, fetchMapStats, subgraphConfigured } from '@/lib/subgraph'
+import { readFeeRateBps } from '@/lib/resaleFee'
 import { logger } from '@/lib/logger'
 import type { MapId } from '@/lib/maps/types'
 
@@ -54,20 +55,6 @@ interface AnalyticsResponse {
 
 // Per-mapId warm-instance cache.
 const cache = new Map<string, { ts: number; value: AnalyticsResponse }>()
-
-async function readFeeRateBps(contractAddress: `0x${string}`, mapId: MapId): Promise<number> {
-  try {
-    const rate = (await fallbackReadClient.readContract({
-      address: contractAddress,
-      abi: MONDETO_ABI,
-      functionName: 'feeRate',
-    })) as bigint
-    return Number(rate)
-  } catch (e) {
-    logger.warn('failed to read feeRate', { err: String(e), mapId })
-    return 0
-  }
-}
 
 const DAY_SECONDS = 86_400
 const WEEK_SECONDS = 604_800
--- a/apps/web/src/app/api/pnl/route.ts
+++ b/apps/web/src/app/api/pnl/route.ts
@@ -3,18 +3,34 @@ import { fallbackReadClient } from '@/lib/chain'
 import { getMapContractById } from '@/lib/maps/contracts'
 import { estimateHistoryFromBlock, scanNormalizedPurchases, toMicrocents } from '@/lib/purchaseLogs'
 import { fetchOwnerPnl, subgraphConfigured } from '@/lib/subgraph'
+import { netOfResaleFee, readFeeRateBps } from '@/lib/resaleFee'
 import { logger } from '@/lib/logger'
 import type { MapId } from '@/lib/maps/types'
 
 /**
  * Server-side profit-and-loss for one wallet on one map: SPENT is the sum of the
- * wallet's own buys; EARNED is what later buyers paid for pixels the wallet used
- * to own. Values are 6-decimal "microcents" (the unit `formatUSDT` renders).
+ * wallet's own buys; EARNED is what actually reached the wallet when later
+ * buyers took pixels off it. Values are 6-decimal "microcents" (the unit
+ * `formatUSDT` renders).
  *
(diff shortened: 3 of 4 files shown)


csacanam:
## Verified end to end against production, not just in unit tests

The PR body says the profile figure wasn't verified end to end. It is now, and without needing a wallet of my own: production runs the old code and this preview runs the new one, so the same address through both is a direct before/after.

Two real wallets off the top of the AREA board:

| wallet | production (gross) | preview (net) | expected at 500 bps |
|---|---|---|---|
| `0x9632…6656` | `88309547` | `83894070` | `83894070` ✅ |
| `0xca117…d643` | `147967171` | `140568813` | `140568813` ✅ |

Both exact. `spent` is byte-identical across the two, which is the control — the buyer does pay the whole price and that number must not move.

This closes three things the unit tests could not:

- **The subgraph path is the live one**, confirming the correction had to sit where it does. Had the fix gone only where #182 pointed, these numbers would be unchanged.
- **The route wires the helper correctly** — the arithmetic being right in isolation didn't prove `computePnl` applies it.
- **`feeRate` reads correctly from the live contract.** 500 bps is what the on-chain value produced here, not a constant in the test.

Worth stating the size in the units players see: those two wallets were being shown **$4.42** and **$7.40** more than actually reached them. On a $147 lifetime figure that is not a rounding complaint.

One limitation unchanged: this exercises `/api/pnl`, not the profile tile that renders it. The tile does no arithmetic of its own, so the remaining risk is display-level only.

GigaHierz (approved):
### Strengths

- **The arithmetic matches the contract exactly.** `apps/contracts/src/Mondeto.sol:206-208` does `uint256 fee = price * _feeRate / 10000; uint256 ownerAmount = price - fee;`. `netOfResaleFee` (`lib/resaleFee.ts:70`) does `gross - (gross * bps) / BPS_DENOM` — fee first, then subtract, same truncation direction. The alternative `gross * 9500n / 10000n` rounds the other way, and the PR knows it: `resaleFee.test.ts` pins `netOfResaleFee(199n, 500) === 190n`, where the other form gives `189n`. That is the single most important detail here and it is correct and test-locked.
- **The rate is read, not hardcoded.** `readFeeRateBps` calls `feeRate()` on the map proxy, and the tests assert rate-independence at 1000 and 2000 bps so nobody can quietly reintroduce a literal 500. Given `MAX_FEE_RATE = 2000` and `setFeeRate` is `onlyOwner`, that is the right posture for a value that has burned this repo before.
- **Verified independently on-chain.** `cast call <proxy> "feeRate()(uint256)"` returns **500 on all eight registered maps** — World, Africa, Asia, Europe, North/South America, Oceania, Antarctica. The issue's premise holds on more maps than it claimed.
- **The fee lands on the right side, and only there.** `pnl/route.ts:66-69` passes `spent` through untouched and nets only `earned`. The buyer pays gross.
- **First sale vs resale is right in both sources and needed no change.** In the log path a pixel with no prior `ownerOf` entry credits nobody; in the subgraph the credit is guarded by `prevOwnerId != ''`. `totalEarned` therefore contains only resale proceeds, which is what makes it safe to net the whole aggregate in one place rather than per-sale.
- **The PR found the bug the issue got wrong.** #182 names the log-scan fallback, which only runs when the subgraph is unconfigured. The subgraph path has the identical defect (`mapping.ts:180` accumulates gross), under a comment claiming it credits "the fee-less per-pixel proceeds" — which was simply false. Fixing only the named path would have left production untouched.
- **No double-count against treasury.** `/api/analytics` computes treasury take as `primaryProceeds + resaleVolume × feeRateBps / 10000`. Sellers' EARNED is now that same fee subtracted, so buyer spend = treasury take + seller earnings closes exactly.

### Issues

#### Critical
None. The money arithmetic is correct.

#### Important

**1. A transient RPC failure silently restores the exact bug this PR fixes — `lib/resaleFee.ts:37-39`.** `readFeeRateBps` returns `0` on a read failure, and `netOfResaleFee(gross, 0)` returns gross. Any Forno hiccup makes the tile show the overstated number again, with only a `logger.warn`. That fallback posture is fine for an ordinary route — it's what `/api/analytics` already did — but this route exists *because* gross is wrong, and there's documented history of region-specific Forno throttling breaking client reads for real users. Fix: memoize the last good rate in module scope (long TTL — it's admin-only, capped, event-emitting, and has never changed on any map) and prefer the cache over `0`.

**2. Adds a serial on-chain round trip to a route that had none on its fast path — `api/pnl/route.ts:54-65`.** Previously the subgraph path was one indexed query and zero RPC calls. Now every cache miss awaits `fetchOwnerPnl`, *then* `readFeeRateBps`, sequentially. `/api/analytics` does the equivalent inside a `Promise.all`; this doesn't. `/api/pnl` is on the profile critical path, and its latency on MiniPay's network is the documented reason it moved server-side. Fix: `Promise.all(...)`, or the cache from (1), which makes it free after the first request. It also fires when `computePnlFromLogs` returns `ZERO` early.

**3. The comment justifying which map's rate is read doesn't hold on the path that serves production — `api/pnl/route.ts:62-64`.** On the subgraph path `fetchOwnerPnl(addr)` returns a cross-map lifetime total (the route says so two lines above), so one map's rate is applied to earnings accumulated across all eight. Harmless today at a uniform 500, but the comment asserts a correspondence that doesn't exist — and reading the rate rather than hardcoding it is precisely an admission that rates can diverge. Reword to state the real assumption so a future divergence is a known trigger rather than a silent wrong number.

#### Minor

**4. Dead import — `api/analytics/route.ts:3`.** `MONDETO_ABI`'s only use was inside the `readFeeRateBps` body this PR deletes. Won't fail CI (`noUnusedLocals` is off, lint isn't gated).

**5. The false subgraph comment survives — `apps/subgraph/src/mapping.ts:229-230`.** Deferring the mapping fix is sound (it needs a full resync), but deleting seven stale words costs nothing and doesn't. Leaving a comment documenting behaviour the code lacks is how the next reader concludes netting is already done and removes the route-level correction.

**6. Pure arithmetic colocated with a server-only module — `lib/resaleFee.ts`.** `netOfResaleFee` is pure and bundle-safe but sits beside `readFeeRateBps`, which imports `@/lib/chain` and `@/lib/logger`. Per `apps/web/CLAUDE.md`, browser code must not import the logger; if anyone ever nets a fee client-side the obvious import drags OTel in. Worth splitting later, not worth churning this PR.

**7. `10_000n` literals remain in `/api/analytics`** while the new helper uses `BPS_DENOM` — the "one definition" goal is half met.

**8. No test asserts the wiring, only the arithmetic.** `resaleFee.test.ts` is genuinely good. What's untested is that `computePnl` nets `earned` and does *not* net `spent`, on both branches. Consistent with the repo (route handlers aren't unit-tested anywhere), so not blocking — but SPENT staying gross is load-bearing with no regression guard.

**9. #182's secondary issue is unaddressed and unmentioned.** The even split `normalized / ids.length` still mis-attributes per-pixel prices across a mixed batch. It doesn't affect the aggregate fee, so it's orthogonal — but Non-goals doesn't name it, and #182 shouldn't be closed as fully done. ("Refs #182" rather than "Closes" is the right call.)

**10. Pre-existing, worth knowing:** the two P&L sources disagree when a wallet re-buys its own pixel. The log scan credits it; the subgraph skips it (`prevOwnerId != buyerId`). On-chain the contract *does* pay `price − fee` there, so the subgraph is wrong. Not introduced here.

### Assessment

**Ready to merge?** **Yes**

**Reasoning:** The fee arithmetic is exactly the contract's, the rate is read on-chain rather than hardcoded, SPENT is untouched, first sales correctly credit nobody, and the expected values are pinned including the truncation boundary. The Important items are hardening and a misleading comment, not wrong numbers — each degrades to today's already-shipped behaviour rather than to a new error.

Two follow-ups worth filing rather than blocking on: memoize + parallelize the rate read (1 and 2), and the subgraph mapping fix — which **must** delete the route-level netting in the same change, or the fee gets applied twice. The PR body flags that itself; worth holding to it, because a resync landing without the route change is a double-deduction on a money tile.

---

*Note: tests were not run as part of this review — no `node_modules` in the reviewing checkout. All seven assertions in `resaleFee.test.ts` were hand-checked arithmetically against the implementation and are correct, including `199n → 190n`, `1n → 1n` and `1_234_567_890_123n → 1_172_839_495_617n`. CI is the real check.*

Issue #201, opened 2026-08-05
Pre-aggregate daily per-wallet pixel gain in the subgraph


GigaHierz:
## Goal

Make "today's campaign board" a **single cheap query** instead of two full paginated block-pinned `ownerMapStats` reads plus a block-by-timestamp binary search. Once the CAMPAIGN tab ships it will be rendered on every `/ranks` visit, so the per-request cost matters.

**Explicitly not a blocker.** Block-pinned diffing is exact and already works today — this is an optimisation, not a dependency of the frontend or admin work.

Context: celo-org/mondeto-admin#51 (admin snapshot + payout).

## Sketch

```graphql
type DailyOwnerMapStat @entity(immutable: false) {
  id: ID!          # "<mapId>-<address>-<dayId>"
  mapId: Int!
  address: Bytes!
  dayId: Int!      # floor(timestamp / 86400), UTC
  pixelsGained: Int!
  pixelsLost: Int!
  netGain: Int!
  spent: BigInt!   # 6-dec microcents
}
```

Updated in `apps/subgraph/src/mapping.ts` on `PixelsPurchased`: `+1` for the `buyer` and `−1` for each `previousOwner`, per pixel — the same bookkeeping `OwnerMapStat` already does, bucketed by UTC day.

## Caveats

- Buckets are **UTC-day-aligned**, so an arbitrary morning→evening campaign window *still* needs the block-pinned diff. This serves the recurring **daily** board, not arbitrary windows.
- Backfill requires a full resync from the start block.
- A future non-gain campaign strategy will want its own bucketed fields — keep the entity easy to extend rather than gain-specific.

## Acceptance

- [ ] Daily `netGain` from the entity equals the block-pinned `ownerMapStats` diff over the same UTC day.
- [ ] Spot-checked on at least one map and two days, including one day with a raid (a wallet with both gains and losses).

GigaHierz:
## Sequencing note: #227 needs the same full resync

Both this issue and #227 require a subgraph resync from the start block, and neither knows about the other.

- Here: *"Backfill requires a full resync from the start block."*
- #227, option A (off-chain price replay, the recommended path): *"Requires a subgraph resync from the start block."*

Two independently-scheduled full resyncs on the same subgraph. Worth batching them — or at minimum landing the `DailyOwnerMapStat` entity in the same mapping change as #227's per-pixel price replay, so one resync serves both.

No change to this issue's scope or its "explicitly not a blocker" status; this is a merge-order note.

[...]

Issue #200, opened 2026-08-05
Board bar: replace the TYCOON tab with a CAMPAIGN tab showing the live campaign leaderboard (UX open)


GigaHierz:
## Goal

The board bar becomes **AREA / EMPIRE / CAMPAIGN**, where the CAMPAIGN tab always shows the currently running campaign's leaderboard. **TYCOON is retired** — it isn't driving anything, and its tab is the one being repurposed.

Right now every board is cumulative-at-a-block, so a one-day campaign just re-crowns the existing long-term holders. The campaign slot fixes that: it ranks what the campaign is actually rewarding, and changes from campaign to campaign.

First campaign strategy: **net pixel gain over the campaign window** — `pixelCount(end) − pixelCount(start)`, growth inside the window, not all-time holdings and not gross purchases. Admin/payout side: celo-org/mondeto-admin#51.

## Ranking — fixed, not open

`netGain` desc → end-of-window `lastGainAt` asc → address asc.

This must match the payout board **byte-for-byte**. That's the lesson from #48, where the payout ranking and `/ranks` had drifted apart.

## Open UX questions — this ticket exists to decide these

1. **Between campaigns** — what does the CAMPAIGN tab show when nothing is running? Hide the tab entirely, show the last completed campaign's result, or fall back to a rolling 24h board?
2. **Window** — the campaign's own `startsAt`→`endsAt`, or a fixed UTC-day board that always exists and campaigns simply attach prizes to?
3. **Timezone** — a "morning to evening" campaign window is local; a UTC day boundary is not. Which does the player see, and how is it labelled?
4. **Tab naming** — the slot is `CAMPAIGN` internally, but players need the *strategy* name ("BIGGEST CLIMB", "TODAY'S GAIN"). Is the tab label per-campaign copy driven by the campaign config?
5. **Negative / zero gain** — hidden entirely, or shown so a player who got raided understands why they dropped?
6. **Row content** — net only (`+12`), or a gained/lost split so the raid mechanic is legible?
7. **Live vs settled** — does the board update live (last-minute sniping is visible, the leader can flip at the buzzer) or freeze at the end of the window?

Question 5/6 matter more than they look: net gain means a wallet raided late in the window **loses ground it had already earned**. That's intended, but it's a genuine surprise unless the UI explains it.

## Out of scope — deliberate, follow-up

[...]

PR #224, opened 2026-08-15
feat(ranks): replace the TYCOONS tab with CAMPAIGN — growth inside the campaign window

csacanam:
~~Draft: this is the data layer only. No tab, nothing a player can see yet.~~ **Outdated — the scope grew after review rounds:** the CAMPAIGN tab, the ranks-page states, the FAQ copy and the settled/settling handling all ship here now. Splitting the ranking out originally was because it is the part that must not drift from the payout, and it is testable on its own — that part still holds.

## Why the board exists

Every board today is cumulative at a single block — LAND, EMPIRE and TYCOONS all answer "who holds the most right now". So a one-day campaign re-crowns whoever was already on top before it started, which is the problem #200 was opened to fix.

This ranks movement *inside* the window instead: `pixelCount(end) − pixelCount(start)`, positive only.

## What's here

**`lib/campaignBoard.ts`** — the diff and a wallet's own standing. 12 tests.

The ordering is **not** reimplemented: it goes through the shared `compareLeaderEntries` (`lib/maps/leaderboards.ts:44`), which already encodes the canonical rule every board uses — value desc → reached-it-first asc → address asc. That is exactly what celo-org/mondeto-admin#51 specifies for the payout, and routing through one comparator is what stops this and the payout snapshot drifting the way they did in #48.

**`lib/subgraph.ts`** — a block-pinned `ownerMapStats` read. The frontend had no way to ask "as of block N" before this; every existing query answers at head.

**`lib/blockAtTimestamp.ts`** — a campaign is scheduled in wall-clock time and the subgraph only pins to blocks, so the boundaries are resolved by binary search rather than estimated. `purchaseLogs.ts` estimates for a scan *start*, where being early is harmless; a window boundary landing a few hundred blocks off silently changes who is in the window.

**`api/campaign-board/route.ts`** — orchestrates the above. Server-side because the two binary searches are ~50 sequential RPC round trips, which has no business in a browser.

## Two decisions worth review

~~**Both boundaries step back ten confirmations.**~~ **Reversed in `cafb596`** (review round 2): the offset made the board's window differ from the payout's, which is the exact drift this PR exists to prevent. Both boundaries now pin with **no offset**; the running board clamps its end to the subgraph head, a settled window is never clamped (`91ceacc`), and reintroducing an offset turns 3 (start) / 4 (end) route tests red.

**No campaign running returns `board: null`, not an error.** Campaigns run on selected days (per the published FAQ), so that is the normal state most days and the client renders its own copy for it.

## Known limit, recorded in the code

`ownStanding` returns the net movement and nothing else. Two pinned `pixelCount` reads **cannot** be decomposed into gains and losses — a wallet ending +12 may have bought twelve and lost none, or bought fifteen and been raided for three, and the two counts are identical either way.

That matters because #200 originally planned to show the split so the raid mechanic is legible. It needs per-event history in the window, which is the `pixelsGained` / `pixelsLost` pair sketched in #201 — currently parked as `priority:low` on the strength of "explicitly not a blocker". True for the ranking; not true for explaining it. Flagged on #200.

## Verification

~~`tsc --noEmit` clean, **437 tests pass** across 53 files.~~ **Stale counts — current head `d53d0ae`: 592 tests across 63 files**, `tsc` clean, coverage floors held, production build green.

**Strongest evidence this PR carries** (added after review round 3): parity was verified against **production data**, not just by reading — the reviewer fetched `ownerMapStats` at the route's own `fromBlock`/`toBlock` straight from Goldsky for the live `campaign-2026-08-21` (330 rows per side), recomputed net gain with the canonical rule, and the preview deployment's response was **byte-for-byte identical**: addresses, values, tiebreaks, order.

The 12 new tests pin the behaviours that decide who gets paid: growth ranked over holdings (a case where LAND's order is exactly reversed), a wallet absent at the start treated as zero, `netGain > 0` only, the `lastGainAt` tie-break, address as the last resort, and **stability across input ordering** — re-running a window must produce an identical order, since the payout is re-derived from it.

**Not verified:** anything end to end. There is no live campaign with `startsAt`/`endsAt` to resolve against, so the route's block resolution and pinned reads are unexercised outside unit tests. That is the next thing to check once the tab exists and a campaign can be scheduled against it.

Refs #200


csacanam:
Out of draft — the tab is wired and the board bar is now **LAND / EMPIRE / CAMPAIGN**.

## What landed since the draft

Narrowing `LeaderboardTab` to drop `TYCOONS` turned the compiler into the checklist: eight errors, one per call site, including the two FAQ answers that name the boards — which `__tests__/app/faq.test.ts` already guards against exactly this kind of drift.

The board arrives through `useCampaignBoard` → `/api/campaign-board` rather than through `useLeaderboard`, because it needs two block-pinned subgraph reads the client's pixel snapshot cannot produce.

The FAQ's "how do leaderboards work" answer now explains *why* the third board is different, not just what it measures — LAND and EMPIRE count holdings so long-time players stay on top, CAMPAIGN counts only movement inside the window so it starts level. That contrast is the whole point of #200 and it belongs in the copy players actually read.

## Sequencing constraint — worth deciding before merge

**This should not ship before celo-org/mondeto-admin#51.** The admin can still configure a campaign to pay TYCOONS today; once that tab is gone from the app, such a campaign would pay a board no player can see. The two sides have to swap together, which is the same class of mismatch as #48 rather than a cosmetic ordering preference.

Happy to hold this until #51 lands, or to merge behind it — but not before.

## Two decisions that hit data walls

Both are recorded on #200 and neither is a blocker; noting them here so a reviewer doesn't look for something that isn't there.

**The gained/lost split isn't rendered.** Two pinned `pixelCount` reads only ever yield the net — a wallet at +12 may have bought twelve or bought fifteen and been raided for three, and the counts are identical. It needs the per-event history sketched in #201. The `gapText` slot stays free, so the split drops in as one line the day that lands.

**"Last campaign" isn't shown between campaigns.** The app only sees the *live* campaign: `readCampaignServer` returns null outside the window, and past campaigns live in the admin-side `campaigns` schedule which the game never reads. So the between-campaigns state says plainly that nothing is running and what the board is for — which was the load-bearing half of that decision anyway. A blank board reads as broken; a stale one reads as live.

## Verification

`tsc --noEmit` clean, **454 tests across 55 files**, coverage gates on both `src/lib/**` and `src/hooks/**` satisfied.

The two new suites pin what actually decides money and comprehension: the block search's at-or-before invariant and its determinism across repeat calls (the payout is re-derived from those blocks), and the hook's handling of a raided wallet that cannot rank but must still be told why.

**Not verified end to end.** There is no live campaign with `startsAt`/`endsAt` to resolve against, so the block resolution and pinned reads remain unexercised outside unit tests. Scheduling a short test campaign against the preview is the cheapest way to close that, and would also be the first real look at the tab with data in it.

csacanam:
## Now verified end to end, on real chain and real subgraph data

The PR said the block resolution and the two pinned reads were unexercised outside unit tests. They aren't any more.

Scheduling a campaign to test this isn't a test — a campaign lives in Edge Config, which every deployment reads, so it would go live for real players on production at the same moment. So the route now accepts an explicit `from`/`to` and honours it **only outside production**, using the same `VERCEL_ENV` gate `app/dev/layout.tsx` already uses. The decision is server-side: the client forwards the params unconditionally and production drops them, so this cannot conjure a campaign for real players.

A 24-hour window against the preview:

```
GET /api/campaign-board?mapId=0&from=2026-08-14T07:06:13Z&to=2026-08-15T07:06:13Z

fromBlock  74790405
toBlock    74876798
entries    [{ address: 0x7de87f…6737, value: 1, tiebreak: 1786725732 }]
```

**The block resolution is auditable from those two numbers alone:** 86,393 blocks across 86,400 seconds is 0.9999 blocks/second, which is Celo's cadence. A linear estimate would have been plausible-looking and quietly wrong; this is the search landing on the right pair.

That exercises the whole chain in one request — timestamp → block binary search on both boundaries, two block-pinned `ownerMapStats` reads, the diff, and the ranking. All of it was unit-tested against synthetic data before; none of it had met the real subgraph.

Visible in the tab too, for anyone who wants to look rather than read JSON:

`/ranks?from=2026-08-14T07:06:13Z&to=2026-08-15T07:06:13Z` on the preview, CAMPAIGN tab. Widen the window for more movement.

**Still unverified:** the path through a genuine Edge Config campaign, since the override bypasses `readCampaignServer`. That is a config read this route shares with the banner, which is already exercised in production — but it is a different line, and worth one look when a real campaign is next scheduled.

GigaHierz (changes requested):
The comparator is the hard part of this and it's done right. The problem is one layer
below it: the *window* handed to that comparator diverges from the payout's.

### Strengths

- **The comparator is genuinely shared, not re-implemented.** `netGainEntries` builds
  `LeaderEntry[]` and hands them to `compareLeaderEntries` (`lib/maps/leaderboards.ts:44`)
  — the same function `fetchAreaLeaderboard` already uses. Compared element by element
  against the admin's `compareEntries`/`rankValues` on `origin/main`, all three keys are
  identical: value desc, `lastGainAt` asc with present-beating-missing, address asc on
  lowercased hex. That is the right structural answer to #48.
- **Address normalisation is correct on both sides** (`campaignBoard.ts:56,61` lowercases
  before the string compare; the admin lowercases in `fetchOwnerMapStats`). No
  checksummed-vs-lowercase ordering hazard.
- **No subgraph dependency added.** `PINNED_OWNER_STATS_QUERY` is `LOCAL_AREA_QUERY` plus
  `block: { number: $block }` — same entity, `where`, `orderBy`, byte-identical to
  admin#51's *Data source* section. No new entity, no `apps/subgraph` change, no
  dependency on #201, exactly as #200 specified.
- **`blockAtTimestamp` matches the admin's semantics** — highest block with timestamp ≤
  target, saturating at head and genesis, including the `<=` on the mid probe.
- **Test quality is high.** 12 lib tests pin the order including a `lastGainAt` tie, an
  address-only tie, absent-at-start, `netGain > 0` only, and input-order stability.

### Issues

#### Critical

**1. `reorgSafe` is applied to the *start* block, which silently shifts the window — and
the admin applies no offset at all. This is the byte-for-byte break.**
`app/api/campaign-board/route.ts` —

```ts
const fromBlock = reorgSafe(fromRaw)
const toBlock   = reorgSafe(toRaw)
```

The window ranked is `[blockAtTimestamp(startsAt) − 10, blockAtTimestamp(endsAt) − 10]`,
not `[startsAt, endsAt]`. On the admin side `resolveBlockNumber`
(`src/lib/contest/snapshot.ts:132`) returns `blockAtTimestamp(t)` with **no offset**, and
admin#51 defines the window as the campaign's own boundaries. I verified both sides
directly.

Two separate defects:
- **The start offset isn't reorg safety at all.** A reorg can only touch blocks near head;
  the start block of a running campaign is hours old. Subtracting 10 there just makes
  ~10s of pre-campaign buys count as in-window gain.
- **The end offset diverges from the payout.** Celo is ~1s/block, so this drops the final
  ~10 seconds of the window — precisely where buzzer-beater sniping happens, and
  precisely the rows that flip a rank.

Any wallet buying in either 10-block band ranks differently on `/ranks` than in the
payout. That is #48, reintroduced one layer below the comparator.

*Fix:* drop `reorgSafe` from `fromBlock` entirely; for the end, clamp rather than offset
(see 2), which gives reorg safety and head safety in one move. If an offset is genuinely
wanted it has to be written into admin#51 and applied identically on both sides — a
unilateral offset on one side is the exact failure mode #200 exists to prevent.

**2. The end block is never clamped to the subgraph's indexed head, so a live campaign
will most likely render "no active campaign".**

`blockAtTimestamp(endSec)` saturates at **chain** head for a future `endsAt`. `reorgSafe`
gives 10 blocks ≈ 10 seconds of headroom against Goldsky's indexing lag, which routinely
exceeds that. The Graph errors on `block: { number: N }` above its head, `querySubgraph`
throws, the `catch` returns `EMPTY`, and the client shows *"no active campaign right
now"* — during a campaign. admin#51 explicitly requires this guard ("keep the existing
`subgraphHead()` guard on the end block"); `lib/subgraph.ts` has no `_meta` query at all.

*Fix:* add `subgraphHead()` (`{ _meta { block { number } } }`) to `lib/subgraph.ts` and
use `toBlock = min(blockAtTimestamp(endsAt), subgraphHead())` — the same construct the
admin already uses, so both sides clamp identically.

#### Important

**3. The board turns on for *any* campaign and ignores `campaign.mapId`.**
`readCampaignServer()` returns the generic Edge Config banner campaign with no
board/strategy field, so a campaign paying the **LAND** board still lights up a CAMPAIGN
tab ranking net pixel gain — a board nobody is paid on. And `CampaignConfig.mapId` is read
nowhere in the route, so a map-3 campaign shows a CAMPAIGN board on all 8 maps. Gate on
the campaign naming CAMPAIGN as a paid board, and return `EMPTY` when
`campaign.mapId != null && campaign.mapId !== mapId`.

**4. The settled board — the one the payout is computed from — is never visible.**
`isCampaignActive` returns `false` the moment `now >= endsAt`, so the board flips to the
between-campaigns state at the buzzer. Winners never see the standings they were paid on,
and there's no way to eyeball the board against the payout. Keep serving the last
completed campaign for a grace period using its own `endsAt` blocks — it's also the
cheapest way to make agreement with the payout observable.

**5. `fromBlock`/`toBlock` are computed, returned, then thrown away.** They're in
`BoardPayload` but `ApiResponse` in `useCampaignBoard.ts` omits them. The one mechanism
that would *guarantee* the payout ranks the same rows — hand the operator the exact pinned
pair and have the payout re-derive from it, which admin#51 already requires via
`fromBlock` — is dropped on the floor.

**6. The "data layer only" framing isn't accurate, and four of #200's seven open UX
questions get answered in passing.** This PR changes `LeaderboardTab` to retire
`TYCOONS`, swaps the tab, rewires `/ranks` and rewrites two FAQ answers — TYCOONS
disappears from the product here. It also decides Q1 (hide between campaigns), Q4 (the
literal string `CAMPAIGN`, which the `BOARD_LABELS` comment itself calls "a placeholder
rather than a decision" — but ships it to players), Q5/Q6 (negative gain hidden, net
only) and Q7 (live, gone at the buzzer). The data layer is flexible enough to express the
alternatives, which is good; the issue is that the player-facing half ships pre-decided.
Either split the UI out and land lib + route alone, or retitle and get those four
answers signed off on #200.

**7. A failed board is indistinguishable from no campaign.** Both the route and the hook
collapse every error into the between-campaigns state. A cache miss costs ~50–60
sequential, unretried `getBlock` calls to Forno plus two paged subgraph reads; one hiccup
and players are told there's no campaign. Also, the start-boundary binary search is
recomputed on every 30s cache miss for every map, though it's invariant once the campaign
starts — resolve the window once per `campaign.id`.

#### Minor

- **`yourNetGain` is plumbed but never rendered** — the raided-player explanation this PR
  argues for exists only in the tab description string.
- **Pagination truncates silently** at `skip <= MAX_SKIP` (6000 rows) where the admin
  loops until a short page. Worse, truncation is by `pixelCount` desc, so it drops the
  *smallest* holders — who at the start block become "absent ⇒ 0", **overstating** their
  net gain. Harmless at today's counts; log when the cap is hit.
- **Dead TYCOONS machinery** — `ranks/page.tsx` still destructures `tycoons`, and
  `/api/global-board` + `useLeaderboard` still compute the whole board nothing renders.
- **`bestBoard` mixes conventions** — `BOARD_LABELS.CAMPAIGN` beside hardcoded
  `'LAND'`/`'EMPIRE'`. The share card's `board` param can now be `"CAMPAIGN"`, which reads
  oddly ("I'm #3 on the CAMPAIGN board").
- **No route-level tests** — both Critical findings live in `route.ts`, the only untested
  file in the change.

### Assessment

**Ready to merge?** **No**

**Reasoning:** The comparator is provably identical to the payout's, which is the hard
part and it's well done — but the window handed to that comparator isn't: `reorgSafe`
shifts both boundaries by ~10s where the admin applies no offset, so the two boards rank
different rows and pay different people. Compounding it, the end block is never clamped to
the subgraph's indexed head, so a live campaign will very likely fail closed to "no active
campaign".

Both fixes are small and local — drop `reorgSafe` on the start, clamp the end to
`subgraphHead()` instead of offsetting it, plus a ~10-line `subgraphHead()` in
`lib/subgraph.ts`. No comparator or test rewrite needed.

---

*Two caveats on this review. admin#51 is not yet implemented on `origin/main` — `BoardKey`
is still `'LAND' | 'EMPIRE' | 'TYCOON'` with no `netGainValues` — so the value/tie-break
comparison is against the issue spec plus the existing `rankValues`/`compareEntries` it
says to reuse. The **block-resolution divergence is against shipped code**
(`resolveBlockNumber`, `snapshot.ts:132`), so that one is solid. And the practical severity
of the head-clamp depends on real Goldsky lag, which I could not measure — but admin#51
requires the guard regardless.*

csacanam:
All seven addressed, and you were right that the window was the real problem — the comparator being shared didn't help while the boundaries it received were someone else's.

## 1 · The offsets are gone

Verified your reading against the admin before changing anything: `resolveBlockNumber` returns `blockAtTimestamp(t)` unmodified. So the ten-block step-back was ranking a window the payout doesn't pay.

Your framing of the start offset is the part I'd got wrong conceptually — it isn't reorg safety at all. A running campaign's start block is hours old and unreorganisable; subtracting from it just let ~10s of pre-campaign buys count as in-window gain.

## 2 · The end clamps to `subgraphHead()`

Added, mirroring the admin's `{ _meta { block { number } } }`. This is the one that would have bitten first: a live campaign's `endsAt` is in the future, `blockAtTimestamp` saturates at *chain* head, The Graph errors on a block it hasn't indexed, and the board renders "no active campaign" **during** a campaign.

Clamping also subsumes the reorg concern the offset was reaching for — indexing necessarily trails chain head.

## 3 · Gated on `campaign.mapId`

A map-3 campaign was lighting a CAMPAIGN board on all eight. The board-name gate isn't possible from here: the public `campaign` key is banner-only by design (the admin README is explicit that tier amounts and boards live in the admin-side `campaigns` schedule). Worth a follow-up to expose the paid board on that key, since a campaign paying LAND still lights this tab.

## 4 · The settled board stays for 48h

`readCampaignForBoard` keeps a finished campaign visible and flags it `settled`, and the tab description says so. Payouts land the day after, so the window has to outlast that — otherwise the ranking a player was paid on is gone before the money arrives.

This also makes the third state real. Running / closed / final was decided on #200, but with the board vanishing at the buzzer only two of them existed.

## 5 · `fromBlock`/`toBlock` reach the client

They were computed, returned and dropped. Now carried through, which is what makes agreement with the payout checkable rather than asserted — admin#51 takes a `fromBlock`, so an operator can re-derive from this exact pair.

## 6 · Retitled

Fair. It said "data layer" because that's what the first commit was, and I never updated it when the UI followed.

On the four UX answers: they're decided and written up on #200 with their reasoning and the existing patterns each reuses, rather than made in passing here. Susanne was asked to lead this and explicitly deferred the two that are hers — the tab label (flagged as a placeholder in `BOARD_LABELS`, one string to change) and the agreement that campaigns must actually be configured on the CAMPAIGN board. Happy to split if you'd still rather, but the decisions aren't unsigned.

## 7 · Failure is distinguishable from no-campaign

Threaded through route, hook and UI. A cold miss is ~50 unretried `getBlock` calls plus two paged subgraph reads, so one hiccup was telling players there is no campaign during one.

## Minors

`yourNetGain` now renders — the unranked row says `RAIDED — −3 PX THIS CAMPAIGN` rather than the generic nudge, which is the raid explanation this PR argued for and didn't show. Dead `tycoons` destructure dropped from `/ranks`.

Left alone: pagination truncation (logging it is a separate change from the `MAX_SKIP` ceiling the AREA board has always had), the global-board/`useLeaderboard` TYCOONS machinery, and `bestBoard`'s mixed conventions.

## Verification

`tsc --noEmit` clean, **484 tests across 59 files**.

The route now has 11 tests where it had none — and **five of them fail against the previous logic**, which is what makes them worth having rather than just present. Both Critical findings lived in that file.

CI can't confirm any of this: the repo-wide workflow failure from #233 means nothing has run since 22:03 yesterday.

GigaHierz (changes requested):
Reviewed at `a2b5309`, rebased onto current main, with the suite and a production build run on the branch.

**Both Critical findings from the previous round are genuinely fixed, and I verified the tests protecting them actually bite** — reintroducing a 10-block offset on `fromBlock` turns `reads the subgraph at exactly those two blocks` red, and only that. The previous round's "no route-level tests" is properly closed: those tests are load-bearing, not decorative.

Verdict: **REQUEST-CHANGES** — one HIGH, one MEDIUM. One round of targeted fixes, not a rethink.

---

## HIGH · `app/api/campaign-board/route.ts:171` — a settled board silently clamps its window, then claims payout authority

```ts
const toBlock = endBlock < head ? endBlock : head
```

The clamp is applied regardless of `settled`. For a **running** campaign it is exactly right, and the test that pins it is right too. For a **settled** one it truncates the window the payout will use — and `ranks/page.tsx:308` then tells the player:

> Campaign closed. This is the ranking the payout settles against.

**The paying side already refuses to do this, deliberately.** `mondeto-admin/src/lib/contest/snapshot.ts:326-332` clamps only for `latest`; given an explicit block or time ahead of the index it throws instead:

> An explicit block/time ahead of the index: never silently re-target an audited block — surface it so the operator picks a valid one.

A campaign window is an explicit time. This route does to it what the payout code documents itself as never doing.

**Failure scenario.** Campaign ends at T. A player opens `/ranks` at T+5s. `blockAtTimestamp` reads the **chain** (`lib/chain.ts` client); `subgraphHead()` reads the **subgraph's indexed head**, which necessarily trails it — so `head < endBlock` immediately after the buzzer is the *expected* state, not an edge case. The board ranks `[start, endBlock − lag]`, dropping the final seconds where buzzer-beater buys land — precisely the rows that flip a rank — and labels the result final. ~30s later the cache expires, the ranking recomputes and changes, after the player was told it was settled.

That is the dispute the code's own comment two lines above says it exists to prevent: *"A player who reads 'closed' as 'final' before the snapshot is taken is where disputes start."*

Verified with a probe against this head: a settled campaign with `head = endBlock − 30` serves `settled: true` and `toBlock = endBlock − 30`. This is money-path checklist item 11 verbatim — *two identical comparators can still disagree when their block windows differ*.

**Fix — one condition.** Mirror the admin's `isLatest`-vs-explicit split: only clamp while the campaign is running. When `settled` and `head < endBlock`, serve a distinct "settling" state (or `EMPTY`) rather than a truncated board wearing the payout's name. The three-state model in decision 4 on #200 already has the vocabulary for it — this is the gap between *closed* and *final*.

## MEDIUM · `__tests__/lib/campaign.test.ts` — flaky, ~1 run in 12

Twelve full-suite runs on this head: **11 pass, 1 fails**. Never fails in isolation. Three *different* cases have failed across runs, always with the same signature:

```
FAIL  campaign.test.ts > readCampaignForBoard > reports a running campaign as not settled
AssertionError: expected null to match object { settled: false }
```

(also seen: `still serves it the day after`, `ignores a finished campaign with no end boundary`)

`readCampaignForBoard` returns `null` only when `EDGE_CONFIG` is unset, `coerceCampaign` rejects, or the dynamic import throws. The helper does `vi.resetModules()` + `vi.stubEnv` + `vi.doMock` *inside* the awaited call, while `afterEach` does `vi.unstubAllEnvs()` + `vi.doUnmock` — and `readCampaignForBoard` resolves `@vercel/edge-config` through a **dynamic** import at call time. That lifecycle is ordering-sensitive; the `expected null` signature is consistent with the env stub being torn down out from under an in-flight call.

I confirmed the symptom rigorously and the mechanism is probable rather than proven.

At 8% this puts main red roughly one run in twelve for no reason — corrosive rather than dangerous, but this repo has just spent a week where trustworthy CI signal mattered a great deal.

**Fix:** move the env stub into `beforeEach`, and replace the `doMock`/`doUnmock` pair with a module-scope `vi.mock` over a mutable ref, so no test's teardown can race another's setup.

---

## Filed as follow-ups rather than blocking this PR

- **#242** — `CampaignConfig` has no `board` field, so the CAMPAIGN tab lights up for a campaign paying LAND. This is also the ops question raised on #200 (*"campaigns must actually be configured on the CAMPAIGN board"*) — it's the same gap, and the config can't express the answer today.
- **#243** — subgraph paging truncates silently at `MAX_SKIP`, dropping the smallest holders, who then read as absent ⇒ 0 at the start block and have their net gain **overstated**.
- **#244** — retired TYCOON machinery still computed on every leaderboard load (16 refs in `useLeaderboard.ts`). Removal is confirmed at product level; kept out of this PR deliberately so the swap and the sweep stay independently revertible.
- **#245** — `bestBoard` hardcodes two labels beside a `BOARD_LABELS` lookup, so the pending CAMPAIGN rename will desync the share card.

## Verified sound — please don't re-litigate these

- **No offset on either boundary**, and the route test catches its reintroduction (proved by reintroducing it).
- **`subgraphHead()` clamp is correct for a running campaign** — the finding above is only about applying it to a settled one.
- `mapId` gating; inverted/empty window rejected with a warn; `toBlock <= fromBlock` fails **closed**.
- `FAILED` distinct from `EMPTY`, with stale-beats-empty on error.
- `fromBlock` / `toBlock` plumbed through to the client, so payout agreement is checkable.
- Settled grace period outlasts next-day payouts (48h).
- Test doubles honour their arguments — the `blockAtTimestamp` fake branches on `sec` rather than returning a constant.
- `windowOverride` is disabled server-side on production `VERCEL_ENV`.
- Production build green, `/api/campaign-board` in the route table; `merge-tree` against main is clean.

## Merge-order note

Shares `lib/subgraph.ts`, `lib/campaign.ts`, `lib/maps/leaderboards.ts`, `ranks/page.tsx` and `faq/content.ts` with main's recent traffic. It is currently rebased and clean — worth merging promptly once the two findings land, rather than letting it go stale again.

## Product

TYCOON retirement is **confirmed**, so the tab swap itself is settled. The remaining open item from #200 is the player-facing **tab label** (question 4) — still shipping as the placeholder `CAMPAIGN`, and it reaches players through the share card as well as the tab bar.

GigaHierz:
Adding one copy change to the review above, and confirming a product decision that removes an open question.

## Copy: "raided" → "bought"

The word doesn't read clearly outside the team, and it's the only thing teaching the mechanic in-app now that the gained/lost split turned out not to be computable. Three player-facing strings, none of which are on `main` yet — so this is a pre-merge fix, not a follow-up:

- `components/Leaderboard/LeaderboardTabs.tsx:50` — `description`
- `components/Leaderboard/LeaderboardTabs.tsx:51` — `globalDescription`
- `app/faq/content.ts:148` — the three-boards answer

```diff
- Who grew the most during the campaign. Pixels you gained, minus pixels raided from you.
+ Who grew the most during the campaign. Pixels you gained, minus pixels bought from you.
```

and in the FAQ, `pixels gained minus pixels raided from you` → `pixels gained minus pixels bought from you`.

"Bought from you" also describes the mechanic more accurately than "raided" does — nothing is taken, it's a purchase at the current price.

Worth aligning the internal comments in the same pass so the vocabulary doesn't fork (`lib/campaignBoard.ts:36,43,78`, `hooks/useCampaignBoard.ts:47,149`, `app/ranks/page.tsx:489`) — comments only, no behaviour.

## Product decisions now settled

Recorded in full on #200; the two that affect this PR:

- **TYCOON retirement is confirmed.** The tab swap is settled, and the dead machinery is #244.
- **Campaigns are not daily.** #200's goal line says "the currently running (daily) campaign"; the live FAQ says campaigns run on selected days. **The FAQ is correct** — building to it, as this PR does, was the right call. The ticket wording is what's wrong.

Still open and still shipping as a placeholder: the player-facing **tab label** (#200 q4). It reaches players through the share card as well as the tab bar, so it's worth settling before this is in front of anyone.

GigaHierz:
One more pre-merge change, closing the last open question on #200.

## Viewer's own row: show `0`, never a negative

`app/ranks/page.tsx:492-495` currently reads:

```tsx
? campaign.yourNetGain < 0
  ? `RAIDED — ${campaign.yourNetGain} PX THIS CAMPAIGN`
  : 'NO GROWTH YET — BUY TO CLIMB'
```

A player at −3 sees **"RAIDED — -3 PX THIS CAMPAIGN"**. The decision is that they should see `0` and still see themselves, so the negative branch collapses into the zero state. That also retires the last player-facing "raided", which lines up with the copy change in my previous comment.

**Clamp at render, not in `ownStanding`.** `lib/campaignBoard.ts:113` should keep returning the true `netGain` — `ranks: netGain > 0` depends on it, and keeping the real value in the data leaves the display decision reversible. One `Math.max(0, …)` at the call site.

Rationale and the accepted tradeoff are on #200; short version: the drop becomes invisible rather than confusing, and the tab description carries the explanation.

---

**All seven of #200's UX questions are now closed.** The only product item still outstanding is the tab label (q4) — this ships as the placeholder `CAMPAIGN` today, and it reaches players through the share card as well as the tab bar.

Running total of what's needed on this PR before merge:

1. **HIGH** — don't clamp a *settled* window to the subgraph head, or don't call it settled (`route.ts:171`)
2. **MEDIUM** — de-flake `campaign.test.ts` (~1 run in 12)
3. **Copy** — "raided" → "bought", 3 player-facing strings
4. **Copy/logic** — viewer's own row shows `0`, not a negative

3 and 4 are small and independent; 1 is the one that matters.

GigaHierz:
Tracing the provenance of `campaignNote` for a scope question, which turned up a defect worth fixing here.

## What it actually is

Not a new UI element. `LeaderboardTabs.tsx:62-65` swaps the **existing per-tab description slot**:

```ts
const activeDescription =
  activeTab === 'CAMPAIGN' && campaignNote
    ? campaignNote
    : active && (scope === 'global' ? active.globalDescription : active.description)
```

CAMPAIGN-only — LAND and EMPIRE keep their descriptions untouched. That's the reuse the #200 answer promised, and it's the right slot.

## The defect: the mechanic's only explanation disappears exactly when it's needed

When `settled` is true, `campaignNote` **replaces** the description. So this:

> Who grew the most during the campaign. Pixels you gained, minus pixels bought from you.

is swapped for this:

> Campaign closed. This is the ranking the payout settles against.

The first sentence is now the **only** in-app explanation of the mechanic — the gained/lost split isn't computable (#200 decision 3, revised), and the viewer's own row will show `0` rather than a negative. So a player who was bought down to zero has exactly one sentence telling them why their number didn't move — and it is removed at the moment they are most likely to be reading the board, i.e. after the campaign closed and the payout is being worked out.

That is the "my count didn't move" support ticket, with the answer deleted from the screen.

**Fix:** render the note *in addition to* the description rather than instead of it — the note is a state line, the description is a rule. They aren't alternatives. Two lines in the same slot, or the note above the tab bar with the description left alone.

## Second: only two of the three states exist

`ranks/page.tsx:306` is the sole assignment, and it's binary — `settled ? '…' : null`. So the shipped model is **closed / not-closed**, not the running / closed / final split the #200 answer argued for.

That matters because the argument for three states was specifically that *"with only two states a player reads the second as final while it can still move."* The implementation has two states, and the second one asserts finality in its copy. That's the HIGH finding in my review, arrived at from the other direction — the design already predicted it.

Resolving the HIGH finding and this are the same piece of work: a settled-but-not-yet-pinned board needs its own state and its own sentence, distinct from a board whose payout block is fixed.

## Scope note

For the record, since this was queried: the three-state model was proposed in the #200 answer on q7 and never explicitly signed off — q7 as written asked only whether the board updates live or freezes. Reusing the description slot is a reasonable reading of that answer, but the extra state was an expansion of the question, not a decision recorded on the ticket. Flagging it as provenance rather than objecting: the direction is right, and the fix above is what makes it true.

csacanam:
Both findings fixed, plus the two product decisions from #200's sign-off that land on this branch.

## HIGH — a settled window is no longer clamped

You were right that the paying side already refuses this, and pointing at `snapshot.ts:326-332` is what made it obvious: the split is `isLatest` vs an explicit pin, and a campaign window is an explicit pin. I had one clamp doing both jobs.

A settled window behind the index now serves a distinct `settling` state — closed, not yet final — rather than a truncated board wearing the payout's name. That is the third state decision 4 asked for, and it turns out the vocabulary was already there; I just hadn't used it at the one point where the two states actually diverge.

Three tests: the settled-and-behind case asserts `settling` **and** that `fetchOwnerStatsAtBlock` was never called, so nothing is ranked from a truncated window; a settled board once the index catches up serves the real `endBlock` unclamped; and `settling` stays distinguishable from both `error` and no-campaign. The running-board clamp and its test are untouched.

## MEDIUM — the flake

Mechanism confirmed as you described. The env stub and module mock were being set inside the awaited call and torn down in `afterEach`, so one test's teardown could race another's in-flight dynamic import — which is exactly why it only ever failed in a full run and never in isolation.

Now a module-scope `vi.mock` over a mutable ref plus a `beforeEach` stub, and `resetModules` is gone entirely.

**Twelve consecutive full-suite runs: 12 green.** Same count you saw one failure in, so this is a like-for-like comparison rather than a smaller sample.

## Product decisions from #200

**q5 — a negative net shows `0`, keeping the row.** Clamped at render with one `Math.max(0, …)`; `ownStanding` still returns the true value, since `ranks` depends on it and a truthful number keeps the display choice reversible — as specified.

**"raided" → "bought"** in the three player-facing strings: the tab description, its global variant, and the FAQ boards answer. The negative branch that carried the fourth use is gone with q5. Internal code comments still say "raided" where they describe the mechanic to a reader rather than a player; happy to sweep those too if you'd rather have one vocabulary.

Also corrected the `(daily)` in this issue's goal line, per the cadence note in your sign-off.

## Verification

`tsc --noEmit` clean, **582 tests across 62 files**, 14 of them on this route.

## Not addressed here

The four follow-ups you filed (#242, #243, #244, #245) stay out, as you scoped them. **q4 — the tab label — is still the placeholder**, and per your cross-link note the order is #224 → #245 → q4, so it isn't this PR's to close.

On your question about @szapelao and future campaigns: answering on #200 rather than burying it here.

GigaHierz (changes requested):
Reviewed at `91ceacc`, rebased on current `main` (0 behind, 9 ahead), with the suite, a production build, and a live comparison against the campaign running today.

**Both findings from the last round are fixed, and the fixes bite.** The settled-window clamp is gone and the `settling` state is served; reintroducing the clamp turns 1 test red, reintroducing a 10-block offset turns 3 (start) / 4 (end) red, deleting `SETTLING` turns 1 red. The flake is gone: **12 consecutive full-suite runs, 12 green** (same sample size as the 1-in-12 failure last round).

**Parity is now verified on production data, not just by reading.** A campaign is live in Edge Config today (`campaign-2026-08-21`, map 0). I fetched `ownerMapStats` at the route's own `fromBlock`/`toBlock` straight from Goldsky (330 rows each side), recomputed net gain with the canonical rule, and compared to what the preview deployment served: **byte-for-byte identical** — addresses, values, tiebreaks, order. Forno confirms block `75398442` has timestamp `1787299200` = `startsAt` to the second and the next block is `+1s`, so "at or before, no offset" is what's running; `toBlock` resolved ~60s behind chain head on a running board, i.e. the subgraph-head clamp is working live.

Verdict: **REQUEST-CHANGES** — one round of targeted fixes, not a rethink. Three MEDIUM, three LOW. The ranking core should not be touched.

---

## MEDIUM · `hooks/useCampaignBoard.ts:74-86`, `:129-133` — the new `settling` state never reaches the player; it renders as "no active campaign"

The route now returns `{ board: null, you: null, settling: true }` for a closed window the index hasn't reached. The hook's `ApiResponse` has no `settling` field and the `!data.board` branch collapses everything non-error into `EMPTY`. `ranks/page.tsx:397` then renders:

> no active campaign right now
> this board ranks growth while one is running

Proved with a probe test: the hook's result for the SETTLING body and for the no-campaign body is `JSON`-identical.

**Failure scenario.** Campaign ends at T. Player opens `/ranks` at T+5s (they were just watching the countdown hit zero). Subgraph trails chain head — ~6 blocks when I measured, routinely more — so the route answers `settling`, and the tab tells them there is no campaign. Thirty seconds later it flips to the settled board with "this is the ranking the payout settles against". The third state exists server-side; the product still has two, and the second one is the wrong one.

**Fix — ~10 lines.** Add `settling?: true` to `ApiResponse` and `settling: boolean` to `CampaignBoardResult`; set it in the `!data.board` branch. In `ranks/page.tsx` add a third empty-state branch before the no-campaign one ("campaign closed — final ranking settling, check back in a moment"). One hook test with the SETTLING body asserting `settling: true` and `failed: false`, alongside the existing no-campaign test as its control.

## MEDIUM · `components/Leaderboard/LeaderboardTabs.tsx:62-65` — the settled note still *replaces* the mechanic's only explanation (raised 2026-08-19, unanswered)

```ts
activeTab === 'CAMPAIGN' && campaignNote ? campaignNote : active && (...description)
```

Unchanged at this head. When `settled` is true, "Pixels you gained, minus pixels bought from you" disappears and "Campaign closed. This is the ranking the payout settles against." takes its slot. With q5 now rendering a negative as `0`, that sentence is the only thing on the screen explaining why a player's number didn't move — and it is removed at exactly the moment they come to read the final board. This was posted as a comment last round and didn't get a FIXED / NOT-FIXED / DISAGREE; flagging it here so it has one.

**Fix — render both.** The note is a *state* line, the description is a *rule*; two lines in the same slot, or the note above the tab bar with the description left alone. No test exists for this component; one render test asserting both strings are present when `campaignNote` is set would make M10 below go red.

## MEDIUM · `app/api/campaign-board/route.ts:107-114` — "ignored entirely on the production deployment" is a guarantee with no test on either path

Mutation: delete the `VERCEL_ENV === 'production'` early-return → **0 tests red** (27/27 green). No route test passes `from`/`to` at all, so neither "production drops it" nor "preview honours it" is pinned. The docblock, the hook's comment ("this cannot conjure a campaign for real players") and last round's response all state the guarantee; nothing would catch a refactor that loses it.

**Fix — two tests.** `vi.stubEnv('VERCEL_ENV', 'production')` + `readCampaignForBoard → null` + `get('mapId=0&from=…&to=…')` → `board` null and `fetchOwnerStatsAtBlock` not called; and the control: unset `VERCEL_ENV`, same request → `board.campaignId` starts with `preview-`.

## LOW · `app/ranks/page.tsx:502-503` — the q5 decision (`0`, never a negative) is pinned nowhere

Mutation: remove `Math.max(0, …)` → **0 tests red**. There is no ranks-page test, so the product decision lives only in a comment. **Fix:** lift the string into a pure helper (e.g. `campaignOwnRowCopy(netGain: number | null)` next to `ownStanding` in `lib/campaignBoard.ts`), call it from the page, and add two unit cases: `-3 → "0 PX THIS CAMPAIGN — BUY TO CLIMB"`, `null → "YOU'RE UNRANKED — CLAIM A PIXEL"`.

## LOW · PR body — three claims are now false, and the body becomes the squash commit on `main`

- Line 1: *"Draft: this is the data layer only. No tab, nothing a player can see yet."* — the tab, the FAQ copy and the ranks page ship here.
- *"Both boundaries step back ten confirmations."* — removed in `cafb596`; the opposite is now the design.
- *"437 tests pass across 53 files"* — 582 / 62 at this head.

Per the rules, strike them through with a pointer to the correcting commit rather than deleting; add a line for the live parity check above if you like, it's the strongest claim this PR can make.

## LOW · `lib/share.ts:45`, `lib/rewards.ts:26` — two doc comments still list `TYCOONS` as a valid board value

Both fields are free strings, so nothing breaks; they just describe a board that no longer exists. One-line edits here, or add both paths to **#244**'s acceptance list — say which.

---

## Verified sound — please don't re-litigate

- `netGainEntries` / `compareLeaderEntries`: unchanged from last round's element-by-element check; now also confirmed against production data (above).
- No offset on either boundary, and the route tests catch reintroduction on both (3 and 4 red).
- Running-board clamp to `subgraphHead()`; settled board unclamped; `settling` served and not cached, so it self-resolves.
- `toBlock <= fromBlock` → `EMPTY` (fails closed); inverted window warned and refused; `netGain <= 0` excluded (mutant → 2 red).
- `mapId` gate (mutant → 1 red); `FAILED` distinct from `EMPTY` (mutant → 1 red); stale-beats-empty on error.
- Test doubles honour their arguments; the settling test's `not.toHaveBeenCalled()` has a live control two tests above it.
- `SETTLED_BOARD_GRACE_MS` = 172,800,000 = 48h; fixture `END_BLOCK − START_BLOCK` = 43,200 = 12h at 1s/block.
- "raided" → "bought" landed in the three player-facing strings; `(daily)` removed from #200's goal line.
- `tsc` clean, 582/582, production build green with `/api/campaign-board` in the route table.

## What the PR got right

The hard part — ranking identically to the payout — is done and is now the best-evidenced piece of code in the repo: shared comparator, no offsets, clamp only where the admin clamps, blocks echoed to the client, and a live byte-for-byte match. Test quality on the route is genuinely load-bearing (7 of 9 mutants die). The two survivors and the client plumbing are all small.

## Merge-order hazards

- **Timing with the live campaign.** `campaign-2026-08-21` runs until 2026-08-22T08:00Z and almost certainly pays LAND — the admin can't target CAMPAIGN until mondeto-admin#51 lands. Merging before it closes puts a CAMPAIGN tab on production for a campaign that pays a different board, and after the buzzer the settled note says "this is the ranking the payout settles against" for a board that isn't. That is #242 verbatim, already filed as a follow-up; the calendar just makes it a day-one event. **Decision: merge Monday 2026-08-24, after the grace window closes at 08:00Z, with no campaign scheduled between 2026-08-22 and the deploy.** Details in the merge-plan comment below.
- Shares `lib/subgraph.ts`, `lib/campaign.ts`, `lib/maps/leaderboards.ts`, `ranks/page.tsx`, `faq/content.ts` with main's recent traffic; currently clean against `8ac29bd`. #242–#245 all rebase onto this.

## Not a finding, for the record

The reply on #200 about future campaigns and @szapelao ("answering on #200 rather than burying it here") hasn't landed yet — #200's last comment is still from the 19th.

GigaHierz:
## Merge plan — agreed today

- **Fix round** for the review findings above lands this week.
- **Merge Monday 2026-08-24**, after the current campaign's settled-board grace window closes: `campaign-2026-08-21` ends 2026-08-22T08:00Z, so the 48h grace ends 2026-08-24T08:00Z. Merging inside that window would put a CAMPAIGN tab — and after the buzzer the "settles against" note — on production for a campaign that pays LAND (the #242 gap, which the admin side can't close until celo-org/mondeto-admin#51 lands).
- @szapelao — this means **no campaign from Sat 2026-08-22 through the Monday deploy (two to three days)**. Scheduling one in that window would light the new tab against a board nobody is paid on. The next campaign can go live as soon as this is deployed; we'll confirm on this thread when it is.

Refs #242, #200.

csacanam:
Round 3 addressed, point by point. All fixes are in `d53d0ae`; every claim below was mutation-verified individually, not inferred from a green suite.

**MEDIUM · settling never reaches the player — FIXED.**
`ApiResponse` carries `settling?: true`, `CampaignBoardResult` carries `settling: boolean`, set in the `!data.board` branch exactly as you sketched. `ranks/page.tsx` renders a third empty state before the no-campaign one: "campaign closed — final ranking settling / check back in a moment". Your probe test is now the pin — the SETTLING body asserts `settling: true, failed: false`, with the no-campaign body as its control two tests below. Mutation: reverting the hook to ignore `settling` turns **1 red**.

**MEDIUM · the settled note replaces the rule — FIXED** (and answering it explicitly this round, since last round it got neither a FIXED nor a DISAGREE: that was a miss, it should have had one).
The note no longer occupies the description's slot. `LeaderboardTabs` renders the note as a state line above the rule, both visible — your framing ("the note is a state line, the description is a rule") is now the comment on the prop. New `LeaderboardTabs.test.tsx` asserts BOTH strings render when `campaignNote` is set, plus two controls (running shows only the rule; another tab never shows the note). Mutation: suppressing the description when the note is present turns **1 red**.

**MEDIUM · the production guarantee has no test on either path — FIXED.**
Two tests, exactly the pair you specified: `VERCEL_ENV=production` + `from`/`to` → `{ board: null, you: null }` AND `fetchOwnerStatsAtBlock` never called; control with the env unset → `board.campaignId` starts with `preview-` and the read happens. Mutation: deleting the early-return now turns **2 red** (your measurement on the review head was 0).

**LOW · q5 pinned nowhere — FIXED.**
`campaignOwnRowCopy(netGain: number | null)` lives next to `ownStanding` in `lib/campaignBoard.ts`; the page calls it. Unit cases: `-3 → "0 PX THIS CAMPAIGN — BUY TO CLIMB"`, `null → "YOU'RE UNRANKED — CLAIM A PIXEL"`, plus `0` and `+4` as controls. Mutation: removing the `Math.max` clamp turns **1 red** (was 0). The clamp stays out of `ownStanding`, which keeps returning the true value, for the reversibility reason the old page comment gave — that reasoning moved into the helper's doc.

**LOW · three false claims in the body — FIXED.**
All three struck through with pointers, per the rules: "data layer only" → corrected to the real scope; "ten confirmations" → points at `cafb596` and states the opposite design plus the 3/4-red guard; "437/53" → corrected to 592/63 at `d53d0ae`. Took your suggestion and added the live parity check as the body's strongest-evidence paragraph.

**LOW · TYCOONS in share.ts/rewards.ts doc comments — FIXED** here rather than deferred to #244: both now name CAMPAIGN and note that older stored records may still carry the TYCOONS string, since both fields are free strings that round-trip persisted data.

Verification on `d53d0ae`:

```
vitest        63 files, 592 tests (582/62 at your review head — +10, all pins above)
tsc --noEmit  clean
coverage      floors held (exit 0)
next build    green, /api/campaign-board in the route table
```

Mutation summary for the new pins: settling 1 · note 1 · prod-gate 2 · q5-clamp 1 — each applied alone and restored before the next.

**On the merge plan:** your decision was "merge Monday 2026-08-24 after the grace window closes at 08:00Z, with no campaign scheduled between the 22nd and the deploy." The time condition has passed. The remaining condition — no campaign currently scheduled in Edge Config — I can't read from here; whoever presses merge should confirm it in the admin first. If one is scheduled, #242's day-one hazard applies and the merge waits for the next gap.

**Still open, honestly:** the reply to @szapelao on #200 about future campaign shapes hasn't landed — it's mine to write and I haven't. Flagging it rather than letting this round imply everything is closed.

GigaHierz (approved):
Reviewed at `d53d0ae`, re-fetched and compared against `headRefOid` immediately before posting (both `d53d0ae…`). Round 4.

## APPROVE — zero open findings

All six from round 3 are fixed, and I verified every one by running it rather than reading the diff. The three that were **0 red** last round now all bite:

| finding | mutation | round 3 | now |
|---|---|---|---|
| M3 production `from`/`to` guarantee | delete the `VERCEL_ENV === 'production'` early-return | **0 red** | **2 red** |
| LOW q5 (`0`, never a negative) | remove `Math.max(0, netGain)` | **0 red** | **1 red** |
| M1 `settling` reaches the player | stop propagating `data.settling` | state absent | **1 red** |
| M2 note beside the rule | make the note replace the description again | no test existed | **1 red** |

Control: 592 passing, 63 files. `tsc` clean, lint 0 errors.

**M1** — the third state is real now, not just served. `useCampaignBoard.ts:95,146` carries `settling` through, and `ranks/page.tsx:391` renders it ahead of the no-campaign branch with the copy the player actually needs. The scenario I described — countdown hits zero, player opens `/ranks`, reads "no active campaign" — no longer happens.

**M2** — `activeNote` and `activeDescription` are now separate values rendered together, so "Pixels you gained, minus pixels bought from you" survives alongside the settled note. That was the one raised in round 2 without a disposition; it has one now, and a render test.

**LOW TYCOONS** — you resolved this better than I proposed. I suggested deleting the references; you kept them and documented *why* the free string can still carry `TYCOONS`:

> `Free string — settled records may carry 'TYCOONS'.`

That is accurate about historical data, where deleting the mention would have made the comment describe a world the records do not live in. Right call.

**LOW body** — all three false claims struck through with the correcting commit named, not overwritten. The `cafb596` pointer on the ten-confirmations reversal is the useful one.

## Behaviour pass

Web build green, with both new surfaces in the route table:

```
├ ƒ /api/campaign-board                    164 B         105 kB
├ ƒ /ranks                                9.4 kB         301 kB
```

**`pnpm run build` at repo root fails — and it is not this PR.** `subgraph#build` dies with `The AssemblyScript compiler crashed when compiling 'src/mapping.ts'`. Separating the signal properly rather than reporting it as a finding:

- this PR changes **0** files under `apps/subgraph`
- swapping `origin/main`'s `apps/subgraph` into the same installed worktree reproduces the identical crash

That is #249, pre-existing. Flagging only so the next reviewer does not spend the twenty minutes I did.

**Merge:** `git merge-tree` against current `main` is clean, and no file this PR touches has been modified on `main` since `8ac29bd` — including through today's traffic (#202, #206, #246, #253 and the pm-kit sync all landed and none of them overlap).

## The one merge-time condition, unchanged in substance

Not a code finding, and #242 already tracks it — but the calendar has moved since round 3, so restating rather than assuming my old date still applies.

`mondeto#242` is **open**, `mondeto-admin#51` is **open**, and `mondeto-admin#78` is **open**. So: the CAMPAIGN tab lights up for *any* running campaign, and the admin still cannot configure a campaign to pay CAMPAIGN. Merged while a campaign is running that pays LAND, the tab shows a board nobody is paid on, directly under the line *"this is the ranking the payout settles against"*.

**I could not verify whether a campaign is live right now** — the public campaign endpoint 404s from here, so I am not going to assert either way. The condition for whoever merges: **confirm no campaign is running, or that any running one pays CAMPAIGN, before pressing the button.** That is a five-second Edge Config check and it is the only thing standing between this and production.

## Verified sound — please don't re-litigate

Everything in round 3's sound list still holds at this head: `netGainEntries` / `compareLeaderEntries` unchanged, no offset on either boundary with reintroduction caught (3 / 4 red), running-board clamp to `subgraphHead()` with a settled window never clamped, `toBlock <= fromBlock` failing closed, `netGain <= 0` excluded, the `mapId` gate, `FAILED` distinct from `EMPTY`, stale-beats-empty on error, and the test doubles honouring their arguments.

## What this round showed

The three findings that mattered were all *guarantees stated in prose with nothing enforcing them* — the production override, the q5 decision, and the settling state. Each is now a mutation that dies. The ranking core has not moved since round 2, which is the right outcome: the part that must not drift from the payout was correct early and stayed untouched while the client plumbing caught up.

PR #228, opened 2026-08-17
fix(wallet): stop unmounting the whole app while the Privy chunk loads

csacanam:
Fixes the five-day browser outage in #221. **The decisive test is the preview** — production throws on every browser load today, so if this preview renders, the mechanism is confirmed and this is the fix.

## The mechanism

`wallet-provider.tsx` rendered `PrivyTree` through `next/dynamic(..., { loading: () => null })`, with `{children}` as that component's child. So on every non-MiniPay load:

1. `mounted === false` → vanilla tree, `{children}` mount under `WagmiProvider` #1
2. `setMounted(true)` → `<PrivyTree>`, chunk not loaded, `loading` renders **`null`** — and since `{children}` is its child, **the entire app tree unmounts**
3. chunk resolves → `{children}` remount under a different provider stack

Step 2 lasts as long as a network fetch. Any async work already in flight in a child resolves during it against refs the unmount had nulled, which is exactly what `Cannot read properties of undefined (reading 'current')` is.

Whether it lands depends on how fast the chunk resolves relative to the children's effects — a **chunk-splitting property**. That explains the three things that made this hard to find: an unrelated postcss bump in #212 flipped it, it was deterministic per build, and no source commit correlated. #209 was suspected and cleared twice.

## The fix

Resolve the import into state instead, so the vanilla tree stays mounted until the component is in hand. The swap becomes one synchronous transition with no interval where children are unmounted.

One remount remains — unavoidable while the vanilla tree and `PrivyTree` each own their own `WagmiProvider` and `QueryClient` — but it is now deterministic rather than a race. Hoisting a single provider spine removes the class of bug and is deliberately left to its own change.

## Also: `PrivyReadyContext` stops asserting readiness

`wallet-provider-privy.tsx` hardcoded `value={true}`, claiming Privy was ready the instant `PrivyProvider` mounted. `usePrivy().ready` is false for a moment after that, and `ConnectButtonInteractive` — gated on this context and itself a second `next/dynamic` boundary — could resolve inside that window and call `useConnectWallet` before Privy had initialised.

That is the likely source of the two `useWallets was called outside the PrivyProvider component` warnings that precede the crash. Driving the context from the real value costs nothing: consumers already treat `false` as "keep the placeholder", which is what those first moments should show anyway.

## Why it's safe

**MiniPay is untouched.** It takes the `isMiniPay` branch, which never enters the Privy path — the `import()` is skipped entirely by the effect's early return. MiniPay has been healthy throughout this outage and must stay that way; nothing here changes its code path.

**No Privy code reaches the server.** A plain `import()` inside `useEffect` runs on the client only, same as `next/dynamic({ ssr: false })` did. `layout.tsx:52` documents why that matters, and the reason is unchanged.

**A failed chunk load degrades instead of crashing.** The `.catch` leaves the client on the vanilla tree: no manual connect button, but a working map. Previously a failure left `{children}` unmounted — a blank page.

`VanillaTree` is extracted rather than duplicated, so the pre-chunk and MiniPay renders are the same component and children genuinely mount once.

## Verification

`tsc --noEmit` clean, **425 tests pass** across 52 files.

**Not verified by any of that** — and this is the point — the outage reproduces only in a real browser. The existing suite passed throughout the five days this was broken, which is itself worth noting: nothing in it exercises the provider swap. `__tests__/components/minipay-privy-isolation.test.ts` guards the bundle boundary, not the mount sequence.

So the test that settles this is loading the preview in a browser. If it renders, confirmed. If it still throws, the unmount is a real bug but not this one, and the next suspect is the `ConnectButtonInteractive` timing above — which this PR also addresses, so a still-broken preview would point somewhere new entirely.

## Follow-ups, not in scope here

- **The single provider spine** (option B in #221) — removes the remount rather than making it deterministic.
- **A regression test for the mount sequence.** Five days of a fully broken browser surface passed CI untouched. Whatever shape it takes, that gap is the real lesson.

Refs #221

### DIFF
--- a/apps/web/src/components/wallet-provider-privy.tsx
+++ b/apps/web/src/components/wallet-provider-privy.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { PrivyProvider } from "@privy-io/react-auth";
+import { PrivyProvider, usePrivy } from "@privy-io/react-auth";
 import {
   WagmiProvider as PrivyWagmiProvider,
   createConfig as createPrivyWagmiConfig,
@@ -40,6 +40,25 @@ const queryClient = new QueryClient();
 // would drag `@privy-io/*` and its `x402` / `@solana/kit` subtree into their
 // chunk. See the note in that module.
 
+/**
+ * Publishes Privy's *actual* readiness rather than asserting it.
+ *
+ * This used to be a hardcoded `value={true}`, which claimed readiness the
+ * instant `PrivyProvider` mounted. `usePrivy().ready` is false for a moment
+ * after that, and `ConnectButtonInteractive` — gated on this context and itself
+ * a second `next/dynamic` boundary — could resolve inside that window and call
+ * `useConnectWallet` before Privy had initialised. That is the likely source of
+ * the two `useWallets was called outside the PrivyProvider component` warnings
+ * that preceded the crash in #221.
+ *
+ * Observing the real value costs nothing: consumers already treat `false` as
+ * "keep the placeholder", which is what the first moments should show anyway.
+ */
+function PrivyReady({ children }: { children: React.ReactNode }) {
+  const { ready } = usePrivy();
+  return <PrivyReadyContext.Provider value={ready}>{children}</PrivyReadyContext.Provider>;
+}
+
 export function PrivyTree({ children }: { children: React.ReactNode }) {
   return (
     <PrivyProvider
@@ -62,11 +81,11 @@ export function PrivyTree({ children }: { children: React.ReactNode }) {
     >
       <QueryClientProvider client={queryClient}>
         <PrivyWagmiProvider config={privyWagmiConfig}>
-          <PrivyReadyContext.Provider value={true}>
+          <PrivyReady>
             <ChainGuard />
             <WalletAnalytics />
             {children}
-          </PrivyReadyContext.Provider>
+          </PrivyReady>
         </PrivyWagmiProvider>
       </QueryClientProvider>
     </PrivyProvider>
--- a/apps/web/src/components/wallet-provider.tsx
+++ b/apps/web/src/components/wallet-provider.tsx
@@ -1,7 +1,6 @@
 "use client";
 
 import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
-import dynamic from "next/dynamic";
 import { useEffect, useState } from "react";
 import { useConnect, WagmiProvider, createConfig } from "wagmi";
 import { injected } from "wagmi/connectors";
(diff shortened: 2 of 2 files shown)


csacanam:
## Confirmed on the preview — the mechanism was the null-gap unmount

The test you described, run: **the preview renders.** Production throws on the same load today, so this is a clean before/after on one code change.

`/` renders the onboarding carousel. `/ranks` renders the full leaderboard. Console filtered for `useWallets|PrivyProvider|Cannot read properties|Application error` returns **nothing** — both warnings and the exception are gone, not just the exception.

**The strongest single signal is the CONNECT button rendering in the top bar.** That is `ConnectButtonInteractive`, which only mounts once `PrivyReadyContext` is true. So the Privy tree loaded, initialised, and published real readiness — the whole path that previously died. If only the crash had been suppressed while the tree stayed broken, that button would still be a placeholder.

So finding #1 in your comment holds: `loading: () => null` unmounting `{children}` for the duration of the chunk fetch was the mechanism, and closing that gap is the fix. Option A was sufficient; the `ConnectButtonInteractive` / `PrivyReadyContext` race was not load-bearing for the crash, though the honest-readiness change stays in on its own merits — asserting `true` at mount was wrong regardless.

## What this doesn't prove

**MiniPay is unexercised here.** It takes the `isMiniPay` branch and skips the `import()` entirely, so by construction nothing changed for it — but "by construction" is an argument, not a device check. MiniPay has been the only healthy surface for five days and is the one that must not regress, so it deserves one look on a handset before this merges. I can do that.

**One remount still happens.** The vanilla tree and `PrivyTree` still own separate `WagmiProvider` and `QueryClient` instances, so children mount, unmount and remount once — deterministically now rather than across a network fetch. That is option B and it is genuinely a different change.

## On the ordering you set out

This unblocks #205, which you were holding until #221 resolved. It touches `layout.tsx`, so it should go in after this rather than alongside.

#220 and #224 are waiting on my changes and stay parked until this is merged.

One thing I'd flag for the follow-up list: **the full suite passed on every one of the five broken days.** Nothing in it exercises the provider swap — `minipay-privy-isolation.test.ts` guards the bundle boundary, not the mount sequence. Whatever shape a regression test takes, that gap is the part worth fixing rather than this specific bug.

csacanam:
**MiniPay confirmed on a handset** — the map loads normally, so the surface that has been healthy throughout this outage is unaffected. That was the last thing standing between this and merge: browser fixed, MiniPay unchanged, both checked rather than argued.

Ready when you are.

csacanam:
Superseded by #229, which shipped the same two fixes and merged while this was in review. Closing rather than leaving a duplicate open.

Both arrived at the same diagnosis and the same shape of fix — resolve the chunk into state so children never unmount, and drive `PrivyReadyContext` from `usePrivy().ready` instead of asserting it. Diffing this branch against merged `main` leaves only naming and comment formatting, so nothing here is missing from what landed.

#229 is the better version: it carries `wallet-provider.test.tsx` and `wallet-provider-privy.test.tsx`, which is exactly the gap I flagged in this PR's follow-ups and didn't close myself. The full suite passed on all five broken days precisely because nothing exercised the mount sequence; that's now covered.

The verification here still stands on its own and I've left it on #221: preview rendering `/` and `/ranks` with a clean console, CONNECT button mounting (so the Privy tree genuinely initialised rather than the crash being swallowed), and MiniPay confirmed unaffected on a handset.

PR #239, opened 2026-08-18
revert: restore self-contained CI — public repos cannot call internal reusable workflows

GigaHierz:
Reverts #233. CI has not run at all since it merged: the run name degraded to `.github/workflows/ci.yml`, meaning GitHub cannot resolve `celo-org/pm-kit/.github/workflows/ci-node.yml@main` — **this repo is public, and GitHub does not allow public repositories to call reusable workflows in internal/private repositories**, regardless of the org Actions-access setting. The other kit repos are private/internal, which is why only this one broke.

Impact while broken: no type-check, no tests, no coverage floor on any branch including `main` since the merge — and the required `ci / ci` check never reports, so merges are hard-blocked.

This revert restores the workflow that was green at the last passing run. The branch ruleset's required check is being updated in lockstep to this workflow's check name. Trade-off: lint gating and the ESLint config from #233 leave with the revert — they return when the repo rejoins the shared CI, which needs a visibility decision (make pm-kit public, or keep this repo's CI self-contained).

Open PRs based on the broken caller (#235, #236) need a rebase onto this once merged before their checks can report.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,11 +1,9 @@
 name: CI
 
-# Gate merges on lint + type-check + unit tests via the pm-kit shared
-# baseline (celo-org/pm-kit ci-node.yml). Deploys/build are handled by
-# Vercel; this workflow is the fast correctness check — in particular the
-# regression guards for the silent RPC-data failures (getLogs chunk size,
-# transport failover config). Coverage floors live in
-# apps/web/vitest.config.ts and fail the `test` step on regression.
+# Gate merges on the web app's type-check + unit tests. Deploys/build are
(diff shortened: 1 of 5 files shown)


GigaHierz:
Closing unmerged — pm-kit is now public, which removes the root cause directly: public repos can call public reusable workflows, and `ci / ci` now resolves and passes on this repo's open PRs (verified on #235 and #236, both green in ~1m35s). What this PR proved and we keep: the restored workflow ran green here (type-check + tests + coverage, 1m11s), confirming the revert path works as a fallback if the visibility decision is ever reversed; the docs caveat lives in pm-kit#14. The branch ruleset's required check has been flipped back to `ci / ci`.

PR #246, opened 2026-08-19
fix(analytics): classify buy failures on every field the error carries

### LINKED ISSUE #215: fix(analytics): pixel_buy_failed hides the root cause — 50.7% of failures are one generic bucket

GigaHierz:
## Problem

Verifying #215's acceptance criteria against the merged code turned up a defect the
issue's own goal covers. #220 gave `pixel_buy_failed` a `category`, but a
**rate-limited RPC failure still lands in `unknown`** — the exact bucket #215
existed to drain.

`useBuyPixels.ts` built its haystack as `` `${msg} ${extractErrorDetail(e)}` ``, and
`extractErrorDetail` returns the single *most specific* field. For viem's
`HttpRequestError` that is `cause.details = "too many requests"`, which matches no
rule in the table. The two fields that **would** match — `cause.shortMessage =
"HTTP request failed."` and the `Status: 429` inside `cause.message` — are
discarded before the classifier ever sees them.

`buyErrors.test.ts` did not catch this because it passes **whole viem envelopes** as
the haystack. The classifier was correct all along; it was being handed a narrower
string than its tests assume. Classic seam bug: both halves tested, the join not.

Reproduced first — the assertion below fails on `main` and passes here:

```
- expected: category: "rpc"
+ received: category: "unknown"
```

## What changes

**`hooks/useBuyPixels.ts`** — new `collectErrorText()` walks the error's `cause` /
`data` chain (depth-capped at 3) and joins every `shortMessage` / `details` /
`message` it finds. `classifyBuy` now gets `` `${hay} ${collectErrorText(e)}` ``.

`extractErrorDetail` is left exactly as it was: "most specific field" is the right
answer for the console line and for the `detail` we send, and only wrong as a
haystack. The two now have separate jobs and say so.

**`isUserRejectedError` deliberately keeps the narrow haystack.** That branch decides
whether the player sees a red error at all or is silently returned to idle; widening
it changes when a failure goes invisible, which is a different risk with a different
blast radius. Out of scope here — flagged, not smuggled in.

**Second fix: the last silent gas path.** Both first-rung `trackGasFallback` calls sat
inside `if (feeCurrency)`, so outside MiniPay a failed estimate only `console.warn`ed
and the transaction went out with **no `gas` field at all** — strictly worse than
either MiniPay rung, and completely invisible. New `GasFallbackLevel` value
`no_gas_limit` covers it. It is disjoint from the other two (no fee currency means no
retry rung), so it must be summed separately rather than folded into the
`without_fee_currency` ⊃ `ceiling` nesting — documented in `analytics.ts`, including
that a `no_gas_limit` carrying `isMiniPay: true` would mean the CIP-64 wiring itself
had broken.

## Tests

The wiring from hook to PostHog was tested at 1 of 4 new emission sites. Now all four,
plus the two guards that had none. **+9 tests, 17 → 24 in this file.**

Every assertion was written first and confirmed red. Mutation results — 7 mutants,
**7 killed, 0 survived**:

| Mutation | Red |
|---|---|
| revert the haystack widening | 1 |
| revert the `no_gas_limit` hoist | 1 |
| drop `category` from the payload | 2 |
| swap `chain_switch_rejected` / `chain_switch_failed` | 2 |
| silence the no-stablecoin guard | 1 |
| report every gas fallback as `stage: 'buy'` | 1 |
| truncate `detail` to 200 instead of 100 | 1 |

Every absence-assertion is paired with a control: the gas-fallback suite has an
explicit "emits nothing on the happy path" test, and the `no_gas_limit` test asserts
`buyCalls()` is length 1 with `gas === undefined` — so it is provably not passing
against a path that never ran. MiniPay is selected via `window.ethereum.isMiniPay`,
the way `getFeeCurrency()` actually reads it, rather than by mocking our own module.

## What this does NOT do

- **No player-facing copy changes.** Every new classification keeps
  `GENERIC_RETRY_MESSAGE`; only the event payload moves. The UX half is still #193.
- **No change to the rejection path** (see above).
- **`no_gas_limit` does not fix the unbounded send** — it makes it countable. Whether
  a non-MiniPay estimate failure should fall back to a ceiling too is a product call
  on gas cost, left for whoever reads the first week of this data.
- **Does not answer #216.** It removes one reason the answer would have been wrong:
  before this, MiniPay rate limits and desktop rate limits both read as `unknown`.

## Verification

```
pnpm run lint       # 22 warnings, unchanged from main (pre-existing, tracked in #210)
pnpm run typecheck  # 2 tasks successful
pnpm run test       # 58 files, 540 tests passed; coverage floors green
```

`useBuyPixels.ts` coverage 95.56% statements / 80.68% branches / 100% functions.
`src/hooks` overall now 44.9% statements (floor is 33).

Not verified automatically and worth one manual pass on the preview deployment:
that a real MiniPay buy still completes, since this touches the CIP-64 gas path —
the emission moved, the estimate and retry behaviour did not.

Closes #215


GigaHierz:
CI green on the head commit: run 32271747179, `ci / ci` pass in 1m36s — lint, typecheck
and test, with the coverage floors in `apps/web/vitest.config.ts` holding. Vercel
preview built and deployed clean.

Restating the one thing CI cannot cover, so it isn't lost on merge: this touches the
CIP-64 gas path, so **a real MiniPay buy should be exercised once on the preview
deployment** before this lands. The emission moved out of the `feeCurrency` guard; the
estimate and retry behaviour did not, and the four-rung suite pins that. But "the mock
agrees" is not the same claim as "MiniPay still signs it".

csacanam (changes requested):
REQUEST-CHANGES — one finding, and it is a missing test rather than a wrong line. The diagnosis and the fix are both right, and I confirmed the fix does what it says.

What I ran, on head 584c1eb:

  pnpm run test
  Test Files  58 passed (58)
  Tests  540 passed (540)

Matches your body exactly.

FINDING — widening the haystack makes rule ordering more load-bearing, and nothing pins that the original six still classify the same way.

RULES is first-match-wins and the file says so in its own words: "the first six rules are the original set and their order is load-bearing." Feeding the classifier strictly more text can only increase the chance an earlier rule claims an error a later rule used to get. That is not a hypothetical failure mode in this repo — it is the documented one. The engineering rules carry it as a worked example: classifier tests asserted on the bare fragment 'execution reverted', real errors carried a docs URL that a different rule matched first, and real reverts were filed as network errors with a fully green suite.

Your mutation table shows exactly 1 test red on "revert the haystack widening." That one test proves the widening does something. Nothing proves it does not also do something else.

I went looking for a live reclassification rather than just asserting the risk. Six constructed viem shapes through the old haystack and the new one, side by side:

     same   rpc                -> rpc                 | rate-limited Forno read (the #215 case)
     same   insufficient_funds -> insufficient_funds  | real ERC20 allowance revert during approve
     same   nonce              -> nonce               | rate limit whose formatted args carry a nonce
     same   permission_denied  -> permission_denied   | MiniPay permission denied on eth_estimateGas
     same   not_land           -> not_land            | NotLand custom-error revert
     same   insufficient_funds -> insufficient_funds  | insufficient CELO for gas, wrapped

Nothing reclassifies. So this is not a bug report — I could not produce one, and I want to be precise about that rather than imply I found something. The marginal risk is also smaller than it first looks, because hay already contained the top-level message and viem repeats most nested text there; what collectErrorText genuinely adds is cause.details and cause.shortMessage, which is exactly the material you are targeting.

The ask is that the probe becomes a test. It is cheap, it is the assertion that would catch the regression this change makes possible, and right now the ordering guarantee is stated in a comment with nothing enforcing it. I have attached what I ran as a starting point; take it or write your own table, but please land one before merging.

Case 3 in that table is worth a second look on its own, and it is NOT caused by this PR. A rate-limited request whose formatted Request Arguments contain "nonce: 42" classifies as nonce, both before and after, because rule 1 wins over rule 9. The player is told "Nonce error — please try again in a few seconds" for what is a rate limit. My fixture is constructed and I have not captured a real viem error proving the nonce token appears in a rate-limited buy, so I am recording this as reachability-unverified rather than confirmed. If you think it is real, it is a follow-up issue, not this PR. If you think it is not reachable, say why and I will drop it.

What I verified as sound:

- The diagnosis is exactly right and better than the issue it closes. extractErrorDetail returns the single most specific field, which for HttpRequestError is cause.details = "too many requests" — matching no rule — while cause.shortMessage and the Status: 429 in cause.message are the two that would match and are discarded. "Both halves tested, the join not" is the correct name for it, and the reproduction with the red assertion is what makes it checkable.
- Leaving extractErrorDetail untouched is the right call. "Most specific field" is correct for the console line and for the detail we send; it was only ever wrong as a haystack. Two jobs, now separated, and the comment says so.
- collectErrorText is bounded: depth cap of 3, and it only walks data and cause. No cycle risk that matters.
- Keeping isUserRejectedError on the narrow haystack is the right restraint, for the reason you give — that branch decides whether the player sees a red error or is silently returned to idle, which is a different blast radius. Flagging it rather than bundling it is what I want to see in a fix PR.
- The no_gas_limit rung is a real second bug and I am glad it is here. Both first-rung trackGasFallback calls sat inside if (feeCurrency), so outside MiniPay a failed estimate only console.warn'd and the transaction went out with no gas field at all — strictly worse than either MiniPay rung and completely invisible. Hoisting the call above the branch is the minimal fix.
- The disjointness reasoning is right and, more importantly, it is written down where the query gets built: no_gas_limit does not nest inside the without_fee_currency ⊃ ceiling relationship, so it must be summed separately. Noting that a no_gas_limit carrying isMiniPay: true would mean the CIP-64 wiring itself broke turns the event into a monitor rather than just a counter.
- Every absence assertion has a control, and the no_gas_limit test asserting buyCalls() length 1 with gas === undefined is the right shape — it proves the send happened AND that it went out unbounded, so it cannot pass against a path that never ran.
- Selecting MiniPay via window.ethereum.isMiniPay, the way getFeeCurrency() actually reads it, rather than mocking your own module, is what makes that suite worth having.

Your "what this does not do" section is accurate and I checked the load-bearing claim: every new classification keeps GENERIC_RETRY_MESSAGE, so no player-facing copy moves. Only the event payload does.

One thing to carry to #251, which I have raised there too: it calls categorizeBuyError(String(err)) on the map-read path. That is the narrow haystack this PR exists to widen, in a sibling path, still open.

GigaHierz:
Pushed `7b7e1ba` (also rebased onto current `main`, which now has #253).

## FINDING — nothing pins that the original six still classify the same way · **FIXED**

Your probe is now a test, and writing it changed my mind about what the assertion should be.

**I started with your framing literally** — classify each envelope through the narrow haystack and the widened one, assert they match. Three of the six failed:

```
× MiniPay permission denied — same category before and after widening
  AssertionError: expected 'permission_denied' to be 'unknown'
× insufficient CELO for gas, wrapped — same category before and after widening
  AssertionError: expected 'insufficient_funds' to be 'unknown'
× rate limit whose args carry a nonce — same category before and after widening
  AssertionError: expected 'nonce' to be 'rpc'
```

The first two are the widening **working** — `unknown → named` is the entire point of this PR. A test demanding nothing changes would have been a test demanding this PR do nothing. So the invariant is narrower than "same category":

> `unknown → named` is allowed and expected. `named → a **different** named category` is the regression — that is what "an earlier rule claimed an error a later rule used to get" actually looks like.

**What landed:** the six envelopes with their categories pinned individually, plus one assertion computing every named→named migration across the table and comparing the whole list. New cases get caught automatically rather than needing a matching assertion each.

**Mutation count: moving the `nonce` rule below `rpc` turns 2 red.** So the guard measures ordering, not just that the rules exist.

Plus a control asserting `collectErrorText` actually returns more text than `String(e)` and that the rate-limit case is rescued from `unknown` — without it, every comparison could be a string against itself and pass against a broken widening.

## On case 3 — you were right to flag it, and right not to claim it

Your third row is real as a mechanism: rule 1 (`nonce`) beats rule 9 (`rpc`), so a rate limit whose formatted Request Arguments contain `nonce: 42` tells the player *"Nonce error — please try again in a few seconds"*. My test reproduces it.

**Reachability in production is still unverified** — the fixture is constructed, and I have not captured a real viem error proving the nonce token appears in a rate-limited buy. So I have not fixed it here, per your own instruction. It is now **pinned as known behaviour** in that migration assertion rather than left as a comment, so if anyone reorders the rules deliberately the test tells them what changed. Filed as a follow-up so it does not evaporate.

## One thing I did beyond the ask — flagging it rather than burying it

`collectErrorText` is **lifted out of `useBuyPixels` into `lib/buyErrors.ts`**, next to the classifier. You suggested this as the fix shape for #251; doing it here is what lets that PR use it without duplicating the walk or depending on a hook. Your parenthetical — *"arguably where it belonged all along"* — is right: it builds the haystack, and the classifier owns the haystack.

Pure move: same function body, same depth cap of 3, same `data`/`cause` walk, now exported and directly unit-testable. Happy to drop it into #251 instead if you would rather keep this PR to the test.

## The sibling path

Confirmed and fixed on that side — #251 now calls the lifted helper instead of `categorizeBuyError(String(err))`. Details in my reply there.

```
$ corepack pnpm run test        # on 7b7e1ba
  Test Files  58 passed (58)
  Tests  548 passed (548)      # was 540
$ corepack pnpm run lint        # 0 errors
$ corepack pnpm run type-check  # 0 errors
```

Ready for re-review.

GigaHierz:
Follow-up for case 3 filed: celo-org/mondeto#261 — kept as **reachability-unverified**, with the first acceptance box being "answer the reachability question", not "write a fix". The constructed fixture stays pinned here as known behaviour so a deliberate reorder shows up as a test change.

csacanam (changes requested):
REQUEST-CHANGES — one finding, and it is two comments in one test file, not code. One commit closes it. Re-reviewed at head 7b7e1ba.

First, what I verified as landed correctly:

- The probe is now a real test with realistic nested envelopes, and your narrowing of the invariant is RIGHT and better than my version: unknown -> named is the point of the PR and must stay allowed; named -> a different named is the regression. That is the correct statement of the property and I am glad my blanket "same category" assertion died on contact with the fixtures.
- Your mutation claim verified by running it: moving the nonce rule below rpc turns exactly 2 red (the case test and the invariant test).
- collectErrorText lifted into lib/buyErrors.ts — answering your question: yes, this is the right home, keep it here and NOT in #251. It sits next to the classifier it feeds, and #251 now imports it from the lib like any other caller. The stacking (#251 retargeted onto this branch) handles the ordering.
- The vacuity control (widened is longer than narrow, and contains 'too many requests') is the right guard for a pair-comparison suite.
- Full suite green at this head. (My local run showed checkSyntaxFloor.test.ts failing — that is my stale node_modules missing acorn after the #253 rebase, not your branch. CI with a fresh install is green. Saying so per the separate-CI-signal rule.)

THE FINDING — two comments in buyErrors.test.ts assert the opposite of what the code next to them proves. I measured both rather than reading them:

(a) The control's last line says "And the widening is what rescues this case from unknown" — but the assertion under it is `expect(categorizeBuyError(narrow(rateLimited))).not.toBe('unknown')`, and it passes because narrow(String(e)) = "Error: HTTP request failed." classifies as rpc, NOT unknown. Measured: narrow -> rpc, widened -> rpc. So for this fixture there is no unknown-rescue; the comment claims the PR's headline behaviour at a site that does not exhibit it. The real rescue is pinned at the hook seam in useBuyPixels.test.ts, which is fine — but this comment will send the next reader looking for a rescue that is not there.

(b) The migration pin's comment says the rpc -> nonce move is "pre-existing, and NOT caused by widening the haystack" — but the test's own data two lines down is before=rpc (narrow), after=nonce (widened). Under the narrow() this file defines, the migration is precisely caused by widening: the nonce token only exists in the nested cause, which only the widened haystack reaches. What IS pre-existing is the rule-order hazard (nonce is rule 1); what widening does is extend that hazard's reach to text that only appears in nested fields — which this fixture demonstrates. The comment says the opposite of what the assertion pins.

Root cause of both: narrow() is String(e), which is not what the pre-fix hook actually passed (message + extractErrorDetail). That's a fine approximation for the invariant test, but the two comments then make claims that are only true of the real pre-fix haystack, not of the narrow() in this file.

Fix, pick either — both are small:

1. Reword the two comments to match the measurements (5 minutes): the control's comment describes what it actually guards (narrow already matches rpc via the outer message; the unknown-rescue is pinned at the hook seam), and the migration comment says "the rule-order hazard is pre-existing; widening extends its reach to nested-only text, as this fixture shows".
2. Or (better, slightly more work): give the rate-limited fixture an outer message that matches no rule — which is the real #215 shape, where the outer error is generic and every classifiable field lives on the cause (exactly the NESTED_RATE_LIMIT_ERROR fixture #251 just built). Then narrow -> unknown, widened -> rpc, the migration list gains 'unknown -> rpc' as the allowed rescue, and both comments become true as written.

Option 2 also makes this file self-contained proof of the PR's headline claim. Either resolves the finding; say which and it's an approve on sight — this is one commit, not a rethink.

Stale prose next to correct code is the exact defect class this repo's rules call worse than silence, which is the only reason this is request-changes rather than a nit.

GigaHierz:
Both comments fixed, and you were right that they were the whole finding. Pushed `04006d8` (fast-forward, so #251's stack is undisturbed).

**Took option 2.** Your diagnosis of the root cause was the useful part: the problem was the fixtures, not the prose. Their *outer* message already carried `HTTP request failed`, so `narrow()` classified them `rpc` and there was no unknown-rescue at this site to point at. Rewording would have made the comments true; changing the fixtures makes the file actually demonstrate the thing.

Both rate-limit envelopes now have a generic outer that matches no rule, with every classifiable field on the cause — the real #215 shape, same as the `NESTED_RATE_LIMIT_ERROR` in #251. Measured:

```
rate-limited Forno read:            narrow -> unknown   widened -> rpc
rate limit carrying a nonce:        narrow -> unknown   widened -> nonce
```

**Three consequences, all improvements:**

**(a) The control now pins the headline claim.** It asserts `narrow -> unknown` and `widened -> rpc` in the same test as the vacuity guard, so the file is self-contained proof of what the PR says it does.

**(b) The migration list is empty** — `toEqual([])`. With the fixtures in their real shape there is no `named -> different named` movement at all, which is the strongest form of the invariant.

**(c) The nonce hazard gets its own assertion**, with the distinction stated the way you put it, because my version of it was simply wrong:

> the rule **order** is pre-existing — `nonce` is rule 1 and tests a bare `includes('nonce')`. What widening changes is the hazard's **reach**: the nonce token here lives only in the cause's formatted Request Arguments, so before widening the classifier never saw it. This fixture is the demonstration.

It also carries a control — strip the nonce from the same envelope and it files as `rpc` — so the assertion pins the rule *order* rather than something about the shape.

**Mutation strength went up, which is the objective evidence option 2 was the better call:**

| mutation | before | after |
|---|---|---|
| move `nonce` below `rpc` | 2 red | 2 red |
| neuter `collectErrorText` (widening does nothing) | **1 red** | **6 red** |

The second row is the point. Previously only the length control noticed; now every fixture depends on the widening, because none of them can be classified without it.

```
$ corepack pnpm run test         # on 04006d8
  Test Files  63 passed (63)
  Tests  629 passed (629)
$ corepack pnpm run lint          # 0 errors
$ corepack pnpm run type-check    # 0 errors
```

On your `checkSyntaxFloor.test.ts` failure — agreed, stale `node_modules` missing `acorn` after the #253 rebase. Mine was green after `pnpm install --frozen-lockfile`. Thanks for separating that from the branch signal rather than reporting it as a finding.

And noted on `collectErrorText` staying here rather than moving to #251 — that settles it.

csacanam (approved):
APPROVE — the finding is closed by 04006d8, and option 2 did what it promised. Verified on this head, locally: full suite 63 files / 629 tests green; neutering collectErrorText turns 6 red where it previously turned 1 (including the hook-seam test), and moving nonce below rpc turns 2 red — both match your table. The two flagged comments are now true as written: narrow -> unknown and widened -> rpc are asserted at the same site as the vacuity control, and the migration list pins toEqual([]).

One nit I measured rather than filed: the order test's control comment says "strip the nonce and the same envelope files as rpc" while the code demonstrates it via CASES[0]; I built CASES[2] minus the nonce token and it does file as rpc, so the prose is true as written and this needs no change.

Carrying your own earlier note to the merge: the one thing CI cannot cover is a real MiniPay buy on the preview deployment — that check is still open and should happen before this lands.

Issue #210, opened 2026-08-10
chore(lint): make lint blocking — clear the 22 warnings and decide on typescript-eslint


GigaHierz:
## Why

`apps/web` is linted, but nothing can fail because of it. `next lint` runs in CI and exits 0 regardless of what it finds, because every rule that fires today fires at `warn` and there is no `--max-warnings` ceiling. So the lint step is currently a report nobody reads, not a gate.

This issue is the decision work: **what makes lint blocking, and which additional rules we run.** (Original title asked which rule set to choose — that is answered, see below.)

## What is already settled — measured, not assumed

- **The rule set is chosen.** `apps/web/.eslintrc.json` extends `next/core-web-vitals` — i.e. the "Strict" option — with a single override, `react/no-unescaped-entities: off`.
- **Lint runs in CI.** `pnpm run lint` → `turbo run lint` → `next lint`, in the required `ci / ci` check.
- **The violation count is known, and it's the small end.** From CI run `32151120943` on `8ac29bd`: **22 warnings, 3 rules, 0 errors.**

| Rule | Count |
|---|---|
| `react-hooks/exhaustive-deps` | 11 |
| `@next/next/no-img-element` | 9 |
| `@next/next/no-page-custom-font` | 2 |

That answers the "measure first" task this issue opened with: it's 22, not 2000, so the "large count → ratchet slowly" branch is not needed. This is fixable in one or two passes.

### 1. Blocking: fix all 22, or ratchet?

At 22 the cheap path is to fix them and gate immediately. Note the three rules are not equivalent work:

- **`react-hooks/exhaustive-deps` (11)** is the one with real bugs behind it. Each is a `useEffect`/`useCallback` whose dependency array omits something it closes over — the class of defect that produces stale reads and effects that don't re-run. These need reading one at a time; some will be genuine fixes, some genuinely want the exhaustive rule suppressed with a reason at the site.
- **`no-img-element` (9)** is a perf recommendation, not a correctness one. On a mobile-first app it is worth taking seriously, but swapping `<img>` for `next/image` changes layout behaviour and is a real diff, not a mechanical one. Turning the rule off with a stated reason is a legitimate answer.
- **`no-page-custom-font` (2)** is a Pages-Router rule firing on App Router code (`src/app/layout.tsx`). Almost certainly a false positive here; verify, then disable with a one-line comment.

### 2. Does `typescript-eslint` go in?

[...]

PR #248, opened 2026-08-19
chore(ci): run CI on staging, move lint off two deprecated tools, align the Node pins

GigaHierz:
Refs #199 — the repo-side half of sections A and B. The evidence table for what in
that issue had already landed is [in the issue thread](https://github.com/celo-org/mondeto/issues/199#issuecomment-5344508625);
this PR is what was actually left that a PR can do.

## Problem

**The lint step runs on two things that are being removed.** CI prints this on every
run:

```
`next lint` is deprecated and will be removed in Next.js 16.
For existing projects, migrate to the ESLint CLI
```

and `eslint@^8.50.0` is end-of-life (npm flags it "no longer supported", no security
fixes). Meanwhile `eslint-config-next@^14.0.0` sat a **full major behind**
`next@^15.5.20`, so the config meant to encode Next's rules was encoding a different
Next's rules.

Two more, smaller: CI ran only on `main` while the documented flow routes every change
through `staging` first, and the Node version had three different answers — root
`engines` `>=18`, `apps/subgraph` `>=20`, `apps/web` no `engines` at all, CI hardcoding
`20`.

## What changes

**ESLint 9 + flat config.** `apps/web/.eslintrc.json` → `apps/web/eslint.config.mjs`,
`lint` script `next lint` → `eslint .`. Flat config rather than eslintrc because it is
the default in ESLint 9 and the *only* supported format in ESLint 10 — writing eslintrc
would have been knowingly scheduling this migration twice.

The version set is forced, not chosen: `eslint-config-next@15` peers
`^7 || ^8 || ^9`, so 9 is the ceiling; ESLint 8 is EOL, so 9 is the floor. ESLint 10 is
`latest` but outside that peer range. `@eslint/eslintrc` is a direct devDependency
because `eslint-config-next` is still an eslintrc-style shareable config and has to come
in through `FlatCompat` — that is also what the upstream `next-lint-to-eslint-cli`
codemod emits.

**CI on `staging`** — both `pull_request` and `push`. A check that only guards the last
hop lets a broken change sit on staging while everyone treats staging as verified.

**`.nvmrc` = 20** and `engines: node >=20.0.0` everywhere. pm-kit's detection prefers
`.nvmrc` over the workflow input, so the input is now a fallback and is commented as one.

**`CLAUDE.md`** — the branch flow, the protection actually enforced by the ruleset, and
a correction: the squash setting is `PR_TITLE` + `PR_BODY`, not `COMMIT_OR_PR_TITLE`,
which is *why* the PR body is the commit message.

## Judgement calls

~~**`run-build` stays `false`, and that is a measurement, not an omission.** #199 asks for
a build gate. I tried it. pm-kit runs `pnpm run build` at the repo **root**, turbo fans
that out to every package, and `subgraph#build` fails on a clean checkout... Flipping the
flag today would turn every PR red... Left off with the reasoning written into `ci.yml`
next to the flag, and filed separately so it outlives this PR.~~

**Struck through — the framing above was wrong, corrected in ecee3d7.** It presented
`run-build: false` as a temporary state blocked on a subgraph fix. It is not: **Vercel
builds, CI is the correctness gate**, and that is the standing org-wide decision rather
than anything local. Verified against the other repos on the pm-kit baseline:

| Repo | Setting |
|---|---|
| `mondeto-admin` | `run-build: false # Vercel builds; matches current gating` |
| `mini-quiz` | `run-build: false` |
| `askbots` | `run-build: false` |
| `saluto` | `run-build: true` (no platform building it) |

pm-kit's own input defaults to `true`, which is right for repos where nothing else builds
the artifact. We have Vercel doing exactly that on every PR, so duplicating it costs merge
latency for a signal we already have. **So #199's build-gate checkbox should be dropped,
not deferred** — the value it was reaching for is already covered.

The `ci.yml` comment now says that, and states the consequence rather than leaving it to
be discovered: a build break surfaces on the **Vercel check**, and only `ci / ci` is a
*required* status check, so a green `ci / ci` is not a claim that the app builds.

**What I found while testing this is still a real bug, just not a blocker.** `pnpm run
build` does fail on a clean checkout — `subgraph#build` runs `graph build` without the
`graph codegen` that writes `generated/`, and `turbo.json` doesn't declare the ordering.
That breaks the documented first command for a new contributor, which is why #249 stays
open — rescoped there to drop the CI-gate framing it was originally filed under.

**Second correction in the same commit:** the `pull_request` trigger no longer enumerates
`[main, staging]`. pm-kit's caller template and two of the three sibling repos leave it
unfiltered, which covers PRs into `staging` without anyone having to remember to edit this
file when a branch is added. `push` stays scoped to `main` and `staging`.

**The lint migration is deliberately behaviour-neutral.** Same `next/core-web-vitals`,
same single `react/no-unescaped-entities` override. Which rules we *want*, and gating on
`--max-warnings`, remain #210 — this PR would otherwise have bundled a rule-set decision
into an infrastructure change.

Two small things the migration itself forced, both visible in the config with a comment
saying why: flat config has no implicit build-output ignores where `next lint` did (so
`.next/`, `coverage/`, `public/` are listed explicitly, or `eslint .` reports on
generated bundles), and the config file is now inside its own lint scope, so the export
is named rather than anonymous — a bare `export default [...]` trips
`import/no-anonymous-default-export` from the very config declaring it.

## Verification

Behaviour-neutrality was checked by diffing findings, not by comparing totals — a count
can match while the contents move:

```
$ diff lint-before.txt lint-after.txt
IDENTICAL — same 22 findings, same files, same rules, same lines
```

Before (`next lint` + eslintrc) and after (`eslint .` + flat config) both:

```
22 problems (0 errors, 22 warnings)
  11  react-hooks/exhaustive-deps
   9  @next/next/no-img-element
   2  @next/next/no-page-custom-font
```

The three commands CI runs, from the root:

```
pnpm run lint       # ✖ 22 problems (0 errors, 22 warnings) — exit 0
pnpm run typecheck  # Tasks: 2 successful, 2 total
pnpm run test       # Test Files 58 passed (58) — coverage floors green
pnpm --filter web build   # builds clean (not gated; see above)
```

**A correction to my own comment on #199:** I reported the warning count there as 26.
It is **22**. The 26 came from grepping a CI log on a feature branch that had added
files, not from `main`. The per-rule breakdown was right; the total was not. This
matters because that number is the input to #210's blocking-vs-ratchet decision, and I
have corrected it there too.

## What this does NOT do

- **No rule changes and no violation cleanup** — all 22 warnings survive untouched (#210).
- **No build gate** (above), and no subgraph or contracts CI.
- **Nothing on `staging` itself.** The workflow will now run on PRs into it, but the
  branch is **175 commits stale** (tip dated 2026-05-26) and has no Vercel domain, so
  the flow this enables is not yet usable end to end. Resetting the branch is a
  force-push to a shared branch and the domain needs Vercel access — both listed as
  owner checkboxes on #199 rather than done here.
- **No `CODEOWNERS`**, no ruleset changes. Section C's remaining items (protect
  `staging`, remove the duplicate repo-level ruleset with its admin bypass) are org
  settings, not repo files.

Not verifiable from here: that the `staging` trigger fires correctly, since no PR into
`staging` exists yet. The syntax is the same list the `main` trigger already uses.


GigaHierz:
CI green on the head commit, and it confirms the migration ran as intended rather than
silently falling back:

```
run 32272603825 — ci / ci  pass  1m31s
✖ 22 problems (0 errors, 22 warnings)
```

Two things to read out of that log. The **`next lint` deprecation notice is gone** — it
printed on every previous run — so the step is genuinely on the ESLint CLI now. And the
count matches the local before/after diff exactly, so the behaviour-neutrality claim
holds on CI's own install, not just mine.

Vercel preview also built clean, which is the check the build gate would have provided
if #249 weren't in the way.

GigaHierz:
## Merge-order hazard: this is a lockfile PR, and it is not the only one

Three other open PRs touch `pnpm-lock.yaml`, all Renovate:

| PR | Title |
|---|---|
| #202 | `chore(deps): update dependency next to v15.5.21 [security]` |
| #206 | `chore(deps): update dependency @privy-io/wagmi to v4.0.16` |
| #89 | `chore(deps): update pnpm to v11 [security]` |

Flagging because this repo has been burned by exactly this twice — the corruption
repaired in #162, and the `rebaseWhen: behind-base-branch` setting added in #166 to stop
it recurring. Renovate only rebases when it next runs, so whichever of these lands
second needs its lockfile **regenerated, not hand-resolved**.

**#202 has a direct interaction with this PR, not just a textual conflict.** It bumps
`next` to 15.5.21 for a security fix; this PR pins `eslint-config-next@^15.5.23`, which
tracks the Next major and is the whole point of the version alignment here. They agree
today — `^15.5.23` and `15.5.21` are the same minor line — but they are two halves of
one version story, so they should not be reviewed as if they were independent.

**Suggested order: #202 first, then this, then #206 and #89.** #202 is a security fix and
the smaller diff; landing it first means this PR rebases onto a tree that already has the
Next bump, and I regenerate the lockfile once rather than Renovate and I taking turns.

Happy to rebase whenever — say the word and I'll regenerate against whatever `main` looks
like at that point. The ruleset's "branch must be up to date" will force the rebase
mechanically either way; the regenerate-don't-hand-resolve part is the bit it can't
enforce.

Worth a look at #89 separately before it merges: it moves pnpm to v11, while this repo
pins `packageManager: pnpm@8.10.0` and `engines.pnpm: >=8.0.0`, and CI installs from that
field. That is a bigger change than its title suggests, and it is unrelated to anything
here.

GigaHierz:
Corrections verified on the head commit `ecee3d7`:

```
ci / ci             pass  1m39s   (event: pull_request)
Vercel – mondeto-web  pass        Deployment has completed
```

The unfiltered `pull_request:` trigger was the one change I could not check locally —
GitHub's own run confirms it parses and fires. Since it carries no branch filter at all,
PRs into `staging` are covered by construction rather than by a list someone has to
maintain. The required check name is unchanged (`ci / ci`), so the ruleset needs no edit.

Ready for review. The two things worth a reviewer's attention are the byte-for-byte lint
parity (that's the whole safety argument for the ESLint 9 migration) and the `ci.yml`
comment now stating the Vercel-builds split as policy plus its cost — a green `ci / ci`
is not a claim that the app builds.

GigaHierz:
## Self-review round — both findings fixed in 4931cea

Reviewed this branch adversarially (running, not reading) and it turned up two things.
Both are fixed; recording them rather than pushing silently.

**1 · `CLAUDE.md` asserted a Node version production doesn't run.** It said `engines`
"agrees with" the pin and left **20** reading as the answer. Vercel runs **24.x** — set on
the project, invisible in the repo, and Vercel does not read `.nvmrc`. Someone chasing a
production-only bug would have built a local repro on the wrong runtime and trusted it.

The pin stays at 20 — I checked whether the *number* was wrong and concluded it isn't
(`mondeto-admin`, `mini-quiz` and `saluto` all pin 20; only pm-kit's template says 24). The
defect was the claim, not the version. The docs now state the gap and what it costs: a green
CI run is not evidence about Node 24 behaviour.

Same reason the **PR title changed** from "one Node answer" to "align the Node pins" — the
repo files now agree, but counting production there are still two, and this title becomes the
commit on `main`.

**2 · The ignores list under-covered gitignored build output.** The body correctly names the
"flat config has no implicit ignores" hazard, then under-applies the fix: `.next/`,
`coverage/`, `public/` were covered, `.vercel/`, `dist/` and `build/` were not — and
`apps/web/.gitignore` exists *solely* for `.vercel`. Proven before and after:

```
before:  gitignored artifacts linted: ['.vercel/output/functions/generated.js', 'dist/bundle.js']
after:   gitignored artifacts linted: NONE
```

Local-only (CI lints a clean checkout), which is exactly why it would have survived a long
time. Anyone running `vercel build` then `pnpm lint` would have linted the whole bundled app.

## Re-verified after the fix

- Lint parity **still byte-for-byte identical** to the pre-migration baseline: same 22
  findings, same files, rules and lines. 208 real files linted, unchanged.
- Gate can still go red — a `react/jsx-key` violation gives `✖ 1 error`, exit 1.
- `pnpm run lint` / `typecheck` / `test` green (58 files); CI green on `4931cea`.

## About the red run in this PR's history

`ci / ci` failed once on this commit and it was **infra, not code**:

```
ETIMEDOUT  request to https://registry.npmjs.org/@types/node/-/node-20.19.43.tgz failed
```

A registry timeout inside `pnpm install --frozen-lockfile`. Proven transient by re-running
the identical commit with no changes — green, install through test.

Worth someone's attention beyond this PR: pm-kit's install step has no retry, so every PR in
every repo on the shared baseline is one network blip away from a spurious red. That is a
pm-kit change and a policy call (retry count, timeout), so I have not filed anything there —
happy to if you want it raised.

## Also verified sound while reviewing (no action needed)

- **No lint coverage regression** — all 200 `src/` files still linted, plus 8 newly covered
  outside `src/`, all clean.
- **No runtime dependency drift** — all 24 of `apps/web`'s production deps resolve to
  identical versions; every added/removed package in the lockfile is ESLint toolchain.
- **`engines.node` does not change Vercel's runtime.** I suspected it might, then refuted it:
  root `package.json` already carried `>=18.0.0` while the project ran 24.x, which proves
  Vercel doesn't resolve ranges to the floor. Narrowing to `>=20.0.0` cannot lower it.

GigaHierz:
Raised the CI flake upstream as **celo-org/pm-kit#18** — the `Install (frozen lockfile)`
step has no step-level retry, so a single registry timeout kills a run in every repo on
the shared baseline.

One correction to what I wrote above, made while writing it up: I said pm-kit's install
step "has no retry". More precisely, it has no **step-level** retry — pnpm already retries
at the fetch layer (`fetchRetries` defaults to 2, exponential backoff), which I verified
against the docs rather than assuming. So that tarball was fetched three times and still
timed out. Fetch-level retry exists and wasn't enough, which is why the fix belongs at the
step.

csacanam (dismissed):
APPROVE. Zero findings. Read the merge-order note at the bottom before you press the button — it is the only thing that can go wrong here, and it is timing, not code.

I checked your claims against the platform rather than against the description.

- The CLAUDE.md correction is right, and it matters more than it looks. Both repos report squash_merge_commit_title: PR_TITLE and squash_merge_commit_message: PR_BODY. The file said COMMIT_OR_PR_TITLE. Your version — "the squash setting is PR_TITLE + PR_BODY, which is why the PR body is the commit message" — is the accurate one, and it is the sentence that explains why everyone writes these long bodies. A doc that got this wrong was quietly undermining the convention it was documenting.
- staging is exactly as stale as you say: tip e83c26f dated 2026-05-26, 175 commits behind main.
- The duplicate ruleset you mention under Section C is real. mondeto has two branch rulesets on main, pm-kit-protect-main and protect-main, both requiring ci / ci. Worth confirming on the record since you flagged it as an org-settings item rather than doing it here.

The behaviour-neutrality check is the part I want to name as the right method. Diffing the findings rather than comparing totals:

  diff lint-before.txt lint-after.txt
  IDENTICAL — same 22 findings, same files, same rules, same lines

A count can match while the contents move. Comparing the contents is what makes "the lint migration is deliberately behaviour-neutral" a claim rather than a hope, and it is why this is a fast approve despite touching the tool that gates every PR.

The version set being forced rather than chosen is worth stating as plainly as you did: eslint-config-next@15 peers ^7 || ^8 || ^9 so 9 is the ceiling, ESLint 8 is EOL so 9 is the floor, ESLint 10 is latest but outside the range. That is not a preference to argue with. (Sidebar: this is exactly why mondeto-admin#68 — eslint to v10 — is red and cannot be rebased into health. Your paragraph here is the diagnosis for that PR too.)

ON YOUR SELF-CORRECTION — this is the strongest thing in the PR and I want it to survive into main.

You struck through the original framing rather than deleting it, which is right, but the substance is what counts: presenting run-build: false as a temporary state blocked on a subgraph fix was wrong, and the corrected version — Vercel builds, CI is the correctness gate, and that is the standing org-wide decision — is supported by the table you built across the four sibling repos. So #199's build-gate checkbox should be dropped, not deferred. Agreed, and I will say so on the issue.

The consequence you wrote into ci.yml is the part nobody would have written unprompted: a build break surfaces on the Vercel check, and only ci / ci is a required status check, so a green ci / ci is not a claim that the app builds. I verified that against the rulesets — ci / ci is the only required context in both rulesets. Stating the limit of your own gate, in the file that implements it, is the difference between a comment and a warning.

Correcting your own 26-to-22 warning count on #199, and saying why it was wrong (grepped a CI log on a feature branch that had added files, not main), is the same instinct. That number feeds #210's blocking-vs-ratchet decision, so getting it wrong quietly would have cost more than the correction did.

The rest, verified as sound:

- Flat config rather than eslintrc, because it is the default in 9 and the only supported format in 10 — writing eslintrc would have been knowingly scheduling this migration twice. Correct reasoning, correctly stated.
- @eslint/eslintrc as a direct devDependency because eslint-config-next is still eslintrc-style and has to come in through FlatCompat, which is also what the upstream codemod emits. Good: the dependency is explained by an upstream fact, not by trial and error.
- The two migration side effects are both in the config with a comment saying why — explicit .next/, coverage/, public/ ignores because flat config has no implicit build-output ignores, and a named export because a bare default trips import/no-anonymous-default-export from the very config declaring it. Both are the kind of thing that looks arbitrary in six months without the comment.
- Holding the rule-set decision and --max-warnings for #210 is the right scope line. All 22 warnings survive untouched, so this is an infrastructure change and nothing else.
- .nvmrc = 20 with the workflow input demoted to a commented fallback, because pm-kit prefers .nvmrc — one answer instead of the three the repo had.

Your "what this does not do" is accurate and honest about the part that is not usable yet: the staging trigger will fire, but the branch is 175 commits stale with no Vercel domain, so the flow this enables is not yet usable end to end. Listing the force-push and the domain as owner checkboxes on #199 rather than doing them here is right — a force-push to a shared branch is not something a PR should smuggle.

MERGE ORDER — the only real risk on this PR.

This touches pnpm-lock.yaml and apps/web/package.json, and #253 touches both as well. I ran the matrix:

  #248 x #253  CONFLICT (content): apps/web/package.json

There are five lockfile-touching PRs open in this repo — #89, #202, #206, #248, #253 — and every pair of them conflicts in pnpm-lock.yaml. Whichever of these lands second must be rebased and have its lockfile REGENERATED, never hand-resolved. That is the rule this repo has been burned by twice.

One ordering fact that is not obvious and decides where this sits: #89 raises packageManager from pnpm@8.10.0 to v11, which rewrites the lockfile format wholesale (lockfileVersion 6.0 to 9.0 — that is why its diff is 9006 lines). Any lockfile work merged before #89 is thrown away when #89 lands; anything after has to be regenerated under pnpm 11 regardless. So #89 needs a decision — land it first or close it and let Renovate re-emit — before this PR or #253 goes in. #89 also conflicts with this one in package.json on engines.node and in ci.yml.

None of that is a defect in this PR. It is what has to be true around it.

GigaHierz:
**Rebased onto `main` after #253 merged — announcing what moved rather than pushing silently.** The approval was dismissed by the push, as the ruleset does on any push.

**One conflict, in `apps/web/package.json`.** #253 added a `browserslist` block in the same place this PR adds `engines`. They are adjacent additions, not competing ones, so both are kept:

```json
"browserslist": ["chrome >= 80", "safari >= 13.1", "firefox >= 74"],
"engines": { "node": ">=20.0.0" },
```

I confirmed #253's `next build && node scripts/check-syntax-floor.mjs` survived the resolution and the guard script is present — that was the thing most at risk from a careless resolve.

**The lockfile was regenerated, not hand-resolved**, per the rule this repo has been burned by twice. Specifically, with the *pinned* pnpm:

```
$ corepack pnpm --version          # honours packageManager: pnpm@8.10.0
8.10.0
$ corepack pnpm install --lockfile-only
$ git status --short
(no change)
$ corepack pnpm install --frozen-lockfile --lockfile-only
Scope: all 3 workspace projects — exit 0
```

Worth stating because it is the trap here: `pnpm` on PATH is **10.24.0**, and regenerating with it would have rewritten `lockfileVersion` 6.0 → 9.0 — silently smuggling #89's format migration into a CI PR. Going through corepack keeps it at **6.0**. Anyone else rebasing one of these five lockfile PRs should use `corepack pnpm`, not `pnpm`.

The regeneration produced no diff, which says the three-way merge was already correct — verified rather than assumed.

**Verification on the rebased head `2596909`:**

```
$ corepack pnpm run lint
✖ 22 problems (0 errors, 22 warnings)

$ corepack pnpm run type-check
tsc --noEmit — 2 tasks successful

$ corepack pnpm run test
Test Files  63 passed (63)
Tests  613 passed (613)
```

The 22 warnings are the same 22 the review diffed as identical before/after — the behaviour-neutrality claim still holds after the rebase, which was the property worth re-checking.

**Merge order from here.** Of the five lockfile PRs, #253 has landed. Remaining: this one, then #202 and #206 (both green, waiting on a Renovate rebase), then **#89 last** — it is the only one whose job is the format rewrite, so it should absorb everything rather than be absorbed, and it is currently red on `web · type-check + tests` independently of any conflict.

Needs a re-approval to merge.

csacanam (dismissed):
APPROVE. Re-reviewed at head 2596909, after the rebase over #253's merge. The rebase is the only thing that changed, so that is what I verified — each claim by looking, not by trusting the description:

- Lockfile format survived: lockfileVersion is '6.0' on this head, same as main. Your corepack warning is well-earned — a PATH pnpm at 10.x regenerating this file would have silently rewritten it to 9.0 and handed the repo a format migration nobody decided. Worth a line in CLAUDE.md at some point ("regenerate with corepack pnpm, never the PATH one"), since the next person hits the same trap; not blocking, this PR's docs scope is already set.
- The conflict resolution kept both sides: apps/web/package.json carries #253's browserslist AND this PR's engines.node >= 20.
- The rebase did not swallow #253's guard: the build script is still `next build && node scripts/check-syntax-floor.mjs` with this PR's lint migration alongside it. That was the riskiest spot in the rebase and it's intact.
- The PR's diff is still exactly its own file set — nothing from #253 leaked in.
- CI green on the rebased head, which per the up-to-date rule is the head that counts.

My original review's approval covered the substance (flat config, forced version set, behaviour-neutral lint diff, the run-build correction); none of that moved, so this round is the mechanical tier: rebase verified, approve renewed.

On the merge-order pushback in your message: conceded — see my reply there. Short version: you're right, #89 goes last. This PR no longer waits on anything.

GigaHierz:
Heads-up, no action needed on this PR and nothing here changes — flagging an interaction so it is not a surprise later.

This adds `.nvmrc` set to `20`, and pm-kit's detection prefers it over the workflow input, so after this merges `.nvmrc` becomes the single source of truth for the CI Node version.

**#89 (pnpm v11) needs Node ≥ 22.13** — `npm view pnpm@11.5.3 engines` returns `{ node: '>=22.13' }`, which is why it fails with `No such built-in module: node:sqlite` on Node 20 before running a single test. So the `.nvmrc` value this PR establishes is the exact thing that has to change for #89 to become mergeable.

Filed as #263, which also picks up the larger point this PR documented rather than closed: CI on 20 while production runs 24, so the version serving players is not the version the suite runs against. Recommending 24 there rather than the 22.13 minimum, for that reason.

**No change requested here.** Introducing `.nvmrc` and keeping it in sync with the workflow fallback is exactly the property #263 needs in order to change one value in one place. Just linking the two so whoever lands #89 does not rediscover it from a red run.

GigaHierz:
Rebuilt on current `main` — announcing rather than pushing silently. Approval was dismissed by the push; content is unchanged from what you approved.

`gh pr update-branch` refused with `Cannot update PR branch due to conflicts` once #202 and #206 landed — both touch `pnpm-lock.yaml`, which is the churn I was holding this PR to avoid. Resolved as a **merge** rather than a rebase (`9060b8f`), so the branch keeps its history and the squash still produces one commit.

**The lockfile was regenerated, not hand-resolved:** took `origin/main`'s copy wholesale, then `corepack pnpm install --lockfile-only` to reconcile it with the merged `package.json`. `corepack` again, not the `pnpm` on PATH — `lockfileVersion` stays `6.0`, and `--frozen-lockfile` passes.

```
$ corepack pnpm run lint          # 0 errors, 22 warnings — the same 22 you diffed
$ corepack pnpm run type-check    # 0 errors
$ corepack pnpm run test
  Test Files  63 passed (63)
  Tests  629 passed (629)
```

**A precise finding about approvals, since we have now measured it both ways today.** #89 is about the ruleset still dismissing on push, but the behaviour is narrower than "any push":

| action | approval |
|---|---|
| `gh pr update-branch` (GitHub's Update-branch button) | **survives** — #202, #206, #246 all kept yours through it |
| any other push, force **or** fast-forward | **dismissed** — this PR just lost it on a plain non-force push |

So GitHub special-cases its own update-branch merge and `dismiss_stale_reviews_on_push` catches everything else. Practical rule until #89 lands: reach for `update-branch` first and only hand-resolve when it refuses — which is exactly when the approval is worth re-earning anyway, because the content genuinely moved.

Needs a re-approve. Next in the agreed order after this is #263 (the Node bump), then #89.

GigaHierz:
**Scope grew by one commit (`87e1cb0`): CI and local move to Node 24.** Flagging it up front because this PR has been read twice already and I have added to it rather than only rebasing.

## Why here rather than a separate PR

This PR already documented the gap in `CLAUDE.md` — CI on 20, Vercel on 24, *"the version serving players is not the version the suite runs on"* — and deliberately scoped to describing it. Two dependencies have since made it load-bearing, so it is now cheaper to close it in the PR that owns the Node story than to merge a doc saying the gap exists and immediately follow with a PR removing it. Its own commit message is "one Node answer".

**Measured, not reasoned:**

```
pnpm 11.5.3   engines: { node: '>=22.13' }
              on Node 20 → "No such built-in module: node:sqlite" (crash before any test)
              on Node 24 → 11.5.3

jsdom 30      engines: { node: '^22.22.2 || ^24.15.0 || >=26' }
              on Node 20 → "webidl.util.markAsUncloneable is not a function", every jsdom worker
              on Node 24 → celo-org/mondeto-admin#69's suite: 43 files, 328 tests, 0 errors, untouched
```

So celo-org/mondeto#89 and celo-org/mondeto-admin#69 are not rebase-able into health; they need this. I installed Node 24 and ran both rather than inferring it from the `engines` fields.

**24, not 22.13.** 22 clears the dependency floor and leaves the production gap open. 24 is what Vercel already runs, so a green CI run becomes evidence about the runtime that serves players — which is the thing this PR set out to make true.

## What moved

Four places, together, because `CLAUDE.md` claims a single Node answer across `packageManager` + `engines` + what CI installs, and leaving any one behind makes that sentence false again:

- `.nvmrc` 20 → 24 — **this is what actually decides**; pm-kit's detection prefers it over the workflow input
- `ci.yml` fallback 20 → 24, kept in sync, with both failure signatures in the comment so the next person does not re-derive them
- `engines.node` `>=20.0.0` → `>=24.0.0` in root and `apps/web`
- `CLAUDE.md` — the "CI and production are not the same version" section rewritten to say they now match, with a paragraph on why it was not always true

**One thing stated rather than implied: `engines` does not enforce.** Control run on Node 20:

```
WARN  Unsupported engine: wanted: {"node":">=24.0.0"} (current: {"node":"v20.19.4"})
exit 0
```

No `engine-strict` in `.npmrc`, and I have not added one — that would abruptly break any contributor on 20 and is a bigger decision than this PR should make. `.nvmrc` is enforcement; `engines` is documentation that warns.

## Verification on Node 24

```
$ corepack pnpm install --frozen-lockfile   # clean
$ corepack pnpm run lint                     # 0 errors, 22 warnings — the same 22
$ corepack pnpm run type-check               # 0 errors
$ corepack pnpm run test
  Test Files  63 passed (63)
  Tests  629 passed (629)
```

Same 22 lint findings as before, so the behaviour-neutrality claim still holds across both the rebase and the Node bump.

**Happy to drop this commit if you would rather keep the PR to what you reviewed** — the CI/staging work stands without it, and #263 can carry the bump separately. It costs one more review round either way; this way it costs zero extra merges.

Still needs a re-approve.

csacanam (approved):
APPROVE — and keep the folded Node 24 commit; you flagged the scope growth, here is the explicit word. Verified: the only addition since the head I approved (2596909) is the merge from main plus 87e1cb0b, which is complete in itself — .nvmrc 20->24 as the source of truth, the workflow fallback kept in sync with the reasoning in a comment, engines bumped in both package.json files, and the CLAUDE.md section this PR itself wrote about the 20-vs-24 gap rewritten in the same diff, so the docs can't drift. ci / ci green on head 87e1cb0 means the whole mondeto suite already ran under Node 24, and 24 is what Vercel serves players with today — the runtime risk is a version production already runs. Splitting the commit out would orphan the CLAUDE.md rewrite and delay #89, which carries a security label. Merge-order note stands: this before #89, and #89 gets the update-branch treatment per your mechanics finding.

Issue #213, opened 2026-08-10
test(wallet): the connect path has no behavioural coverage


GigaHierz:
## Why

The wallet connect path has **no behavioural test coverage at all**, and it is the path every player passes through before they can spend money.

What exists today:

- `TopBar.test.tsx` renders `ConnectButton` — but `vi.mock`s it into `() => <button>Connect</button>`. It asserts the bar renders a stub, not that the wallet works.
- Four test files `vi.mock('wagmi')`, so the connector layer is never exercised.
- `minipay-privy-isolation.test.ts` (#209) is static source analysis. It guards the import boundary; it asserts nothing about behaviour.

Nothing else imports Privy or renders a wallet provider tree.

This came up while reviewing the open PRs: the two touching wallet code (#209, #206) were the two with no automated safety net, and both had to be validated by a human clicking through a preview. That is the entire test suite for this path.

## What to cover

Roughly in order of value:

- [ ] **MiniPay branch** — `window.ethereum.isMiniPay === true` renders the vanilla wagmi tree, auto-connects the injected connector, and never mounts `PrivyTree`. This is the majority path and the one where a regression is most expensive.
- [ ] **`ConnectButton` states** — disconnected shows CONNECT; connected shows the profile label; falls back to `generateUsername` when there is no on-chain label; returns `null` in MiniPay.
- [ ] **Connected menu** — opens on click, PROFILE links to `/profile`, LOG OUT calls `disconnect`, outside-click closes it.
- [ ] **SSR / first-paint parity** — the `mounted` gate renders the same placeholder on server and first client render. This one has bitten before: the architecture comment in `wallet-provider.tsx` records that a hydration mismatch previously crashed the lazy Privy chunk mid-load.
- [ ] **Desktop branch** — non-MiniPay mounts `PrivyTree`, `PrivyReadyContext` flips to `true`, and the interactive button replaces the placeholder.

## How

`@testing-library/react` is already set up and the suite has 12 component tests to follow. The wagmi/Privy providers need mocking at the module boundary rather than avoiding the components entirely — the current approach mocks the component under test, which is why the coverage is illusory.

Non-goal: testing Privy's own SDK. The contract worth pinning is *ours* — which tree renders under which condition, and what the button does in each state.

## Acceptance

A regression in the connect flow fails CI without anyone opening a browser. Concretely: reverting the module split from #209, or breaking the MiniPay branch in `wallet-provider.tsx`, turns the suite red.

[...]

PR #250, opened 2026-08-19
test(wallet): cover the connect path players actually use

GigaHierz:
## Problem

`ConnectButton` is the only surface a signed-out player can act on, and the only
route to LOG OUT for a signed-in one. It had **zero** behavioural coverage.

Verified against `main` before writing anything, because #213 was filed a while ago
and parts of it have since been overtaken:

- `TopBar.test.tsx:5` still does
  `vi.mock('@/components/connect-button', () => ({ ConnectButton: () => <button>Connect</button> }))`
  — a stub that contributes coverage while asserting nothing about the real component.
- **5** test files `vi.mock('wagmi')` (the issue said 4), so the connected/disconnected
  split was never exercised.
- `minipay-privy-isolation.test.ts` is static source analysis. It guards the import
  boundary from #209; it asserts nothing about behaviour.

## Scope — two of the issue's five boxes were already done

#229 landed `wallet-provider.test.tsx` and `wallet-provider-privy.test.tsx` after #213
was filed, which covers the **MiniPay provider branch** and the **desktop /
PrivyReadyContext branch**. I confirmed those aren't vacuous rather than assuming:
forcing the MiniPay detection off in `wallet-provider.tsx:70` turns **1** test red.

So this PR covers the three boxes that were genuinely open — `ConnectButton` states,
the connected menu, and first-paint parity — instead of re-doing work that exists.

## What changes

Two new test files, no product code touched.

**`connect-button.test.tsx`** — the shell's three-way decision on every page load:
placeholder before the Privy chunk is ready, `null` in MiniPay, interactive half once
ready. The MiniPay case is asserted with `PrivyReadyContext value={true}` on purpose —
that tests the *ordering* of the branches, not just the flag, since a MiniPay user must
never be offered a connect action that cannot succeed.

The interactive half is stubbed **in this file only**, deliberately: it is loaded via
`next/dynamic({ ssr: false })` precisely so `@privy-io/react-auth` stays out of the
shared chunk, and importing it for real here would have the test reintroduce the
coupling #209 removed.

**`connect-button-interactive.test.tsx`** — label resolution (on-chain name → generated
username → `…`), connect vs. menu-toggle branching, and the menu: PROFILE href, LOG OUT
calling `disconnect`, outside-click dismissal, inside-click persistence.

The generated-username expectation is recomputed with the real `generateUsername` rather
than hard-coded, so a change to the word lists doesn't fail a test that isn't about them.

## Verification

Every "it closed / it didn't fire" assertion is paired with a control proving the
opposite state is reachable, and the outside-click test fires `mouseDown` rather than
`click` — firing `click` would pass against a component that registered no listener at
all.

**Nine mutations, nine caught, zero survived:**

| Mutation | Tests red |
|---|---|
| swap the connect / menu-toggle branches | 6 |
| invert the `privyReady` gate | 4 |
| label always reads `CONNECT` | 3 |
| invert the outside-click containment check | 2 |
| drop the MiniPay early return | 1 |
| drop the on-chain-name preference | 1 |
| LOG OUT no longer calls `disconnect` | 1 |
| listen on `click` instead of `mousedown` | 1 |
| PROFILE href → `/` | 1 |

```
pnpm run lint       # 22 problems (0 errors) — unchanged
pnpm run typecheck  # Tasks: 2 successful, 2 total
pnpm run test       # Test Files 60 passed (60) — was 58
```

Coverage: `connect-button.tsx` **0 → 100%** statements/branches,
`connect-button-interactive.tsx` **0 → 96.42%** statements, **100%** branches.

## What this does NOT do

- **`TopBar.test.tsx` keeps its stub.** #213 calls that mock out as the reason coverage
  was illusory, and it was — but the fix is that *something* tests the real component,
  which now happens here. Rendering the real one inside TopBar's suite would drag Privy
  and wagmi into a test about a layout bar. Flagging it as a deliberate call rather than
  an oversight.
- **No product code changes**, so nothing here alters what a player sees.
- Does not test Privy's own SDK — the contract worth pinning is ours.

## Worth a separate look

The coverage floors in `apps/web/vitest.config.ts` now sit **~8 points below actual**
(floors 28/25/28/29 vs 36.35/32.28/36.87/36.86), because #205, #236 and this PR all
raised coverage without moving them. The config's own comment says they "are meant to
ratchet, and they only ratchet if someone moves them". I left them alone here rather
than widen this PR, and because raising them mid-flight would put the other open PRs at
risk of a spurious red. Happy to do it as a one-liner once the queue drains.

Refs #213


csacanam (approved):
Reviewed at head 56945f3, re-fetched and compared against headRefOid immediately before posting.

APPROVE. Zero findings. Tests only, no product code, and the test hygiene is the reason this is a fast approve rather than a slow one.

What I verified:

  pnpm run test    # on 584c1eb-era main + this branch
  Test Files  60 passed (60)

I checked the specific claims that are easy to make and hard to keep, rather than taking the mutation table on trust.

- The outside-click test fires mouseDown, not click. Confirmed at connect-button-interactive.test.tsx. This is the difference between a real assertion and one that passes against a component that registered no listener at all, and you called that out in the body before anyone asked.
- Every absence assertion has a control next to it. "stays open when the click lands inside the menu" is the control for the outside-click case; "still renders the button for a non-MiniPay injected wallet (control)" is the control for the MiniPay early return. Both are the pairing that stops "nothing happened" passing against code that never ran.
- The generated-username expectation is recomputed with the real generateUsername, and then guarded with expect(expected).not.toBe(''). That second line is the part I want to name: without it, a change that made generateUsername return empty would turn the assertion into empty-equals-empty and stay green. Recomputing instead of hard-coding is good; proving the recomputed value is non-empty is what makes it non-vacuous.

On your three judgement calls, all correct:

- Stubbing the interactive half in connect-button.test.tsx only. The component is loaded via next/dynamic({ ssr: false }) precisely so @privy-io/react-auth stays out of the shared chunk; importing it for real would have the test reintroduce the coupling #209 removed. Right call, and scoped to the one file that needs it.
- Asserting the MiniPay case with PrivyReadyContext value={true}. That tests the ordering of the branches rather than the flag, which is the property that matters: a MiniPay user must never be offered a connect action that cannot succeed. A test that only set the flag would pass against code with the branches swapped.
- Leaving TopBar.test.tsx's stub in place. #213 named that mock as the reason coverage was illusory and it was — but the fix is that something tests the real component, which now happens here. Dragging Privy and wagmi into a test about a layout bar to satisfy the letter of the issue would have made the suite worse.

Scope discipline: verifying against main before writing, finding that #229 had already covered two of #213's five boxes, and confirming those weren't vacuous by forcing the MiniPay detection off in wallet-provider.tsx:70 to see 1 go red — that is the right way to avoid re-doing landed work. Correcting the issue's "4 test files vi.mock('wagmi')" to 5 in the same pass is the kind of small record correction that costs nothing and stops the next person re-deriving it.

Refs #213 rather than Closes is correct: three of five boxes.

ON THE COVERAGE FLOORS — agreed with your call to leave them, and I want to make sure the follow-up actually gets filed rather than living in this PR body.

You are right that the floors now sit ~8 points below actual (28/25/28/29 vs 36.35/32.28/36.87/36.86) and that raising them mid-flight would put the other open PRs at spurious risk. Do it as the one-liner once the queue drains, as you say.

But the config's own comment says the floors "are meant to ratchet, and they only ratchet if someone moves them," and this is the third PR in a row (#205, #236, this) to raise coverage without moving them. That pattern is exactly how a ratchet quietly stops being one. Please file it as an issue now, so it survives this PR being merged and closed:

  Title: Coverage floors have drifted ~8 points below actual — ratchet them
  Labels: priority:low, size:S

  apps/web/vitest.config.ts floors are 28/25/28/29; actual after #205, #236 and
  #250 is 36.35/32.28/36.87/36.86. The config comment says the floors are meant
  to ratchet and only ratchet if someone moves them; three PRs have raised
  coverage without moving them.

  Deliberately not done inside #250: raising them mid-flight risks a spurious
  red on the other open PRs. Do it once the current queue drains.

  Refs #213.

Not blocking the approve — the PR is right to leave it alone. Just do not let it evaporate with the PR body.

Issue #226, opened 2026-08-15
feat(analytics): instrument map mount so a freeze on entry is visible


GigaHierz:
## Why

#196 reports MiniPay hanging for minutes on entry and then force-closing. We cannot
currently tell whether that is fixed, getting worse, or how many players it affects,
because **a client that freezes or fails to parse never fires a single PostHog event**.
The affected population is structurally absent from client analytics.

That gap is why #196 sat on two support reports as its entire evidence base, and why we
still can't answer "is it two players or two thousand?" — or, more practically, "can we
ask the reporter to test again yet?"

This was item 4 of #196's own suggested investigation and was never built.

## What to add

A mount lifecycle event pair on the map, the heaviest client path:

- `map_mount_started` — fired as early as possible in the client render, before the
  full-grid pixel read and land-mask decode.
- `map_mount_completed` — fired once the grid is painted and interactive, carrying the
  elapsed time.

The ratio `completed / started` is the freeze rate. The elapsed-time distribution shows how
close the slow tail runs to the host's not-responding watchdog, which is the thing that
actually force-closes MiniPay.

Worth carrying on both: `isMiniPay` (already a PostHog super-property, per
`lib/analytics.ts`), the map id, and whether the read was served from cache.

## Why the pair, and not just one event

A single "map loaded" event can't distinguish *never started* from *started and hung* —
and those have completely different causes. A parse failure produces neither event; a
main-thread freeze produces `started` and no `completed`. Being able to tell those apart is
the whole point, because #196 currently conflates them:

- the **freeze / ANR** symptom (multi-minute hang, "not responding" dialog) points at the
  mount read blocking the main thread
- the **blank map** symptom points at the bundle failing to parse on an old WebView

The parse-error path is tracked separately. This ticket is what makes the freeze path
measurable, and what tells the two apart in the data instead of by guessing from a
device model.

## Complements, doesn't duplicate, #205

#205 counts *engines* server-side at the document request, which catches clients whose
bundle never parses — they can't report anything themselves. This ticket covers the
clients that **do** boot and then hang, which #205 cannot see because they look identical
to a healthy client at the document request.

Together they give a complete funnel: document request → bundle parsed → map mount started
→ map mount completed. Each drop-off has a distinct cause.

## Acceptance

- [ ] Both events fire on a healthy load, in order, with elapsed time on the second.
- [ ] A deliberately stalled chain read produces `started` with no `completed`.
[...]

Issue #225, opened 2026-08-15
fix(minipay): down-level dependency syntax to the Chrome 80 support floor


GigaHierz:
## Why this exists

The support floor is **Chromium 80** (decided in #196). Our client bundle currently
cannot be *parsed* by an engine at that floor, so affected devices show a blank map and
never run a line of our code — including PostHog, which is why the population is
invisible in analytics.

This work has had no ticket of its own. It was investigated in draft PR #204 and
partially reduced by #209, but the remaining half is tracked nowhere, and #196 was closed
in the meantime. This is its home.

## What's actually left

#209 removed the Privy / `x402` / `@solana/kit` subtree from the client path, which cut
logical assignment operators (`||=`, `??=` — Chrome 85+) on the route from **21 to 6**.
It did not finish the job, and says so itself:

> It does not close that issue — viem, @noble/hashes and @wagmi/connectors still ship
> syntax above the Chrome 80 support floor, and all three are genuinely required.

So the remaining sources are three packages that cannot simply be dropped:

| Package | On the MiniPay path? |
|---|---|
| `viem` | yes — unavoidable |
| `@noble/hashes` | yes — transitive dependency of `viem` |
| `@wagmi/connectors` | yes |

Six occurrences is still a parse error. A parse error kills the whole chunk before
anything inside it runs.

## The trap to avoid — recorded so it isn't rediscovered

PR #204's body contains a "what was tried and did not work" table whose
`transpilePackages` row is **misleading**. It reports 29 occurrences surviving after
naming five packages, and concludes the mechanism doesn't work. It does work; it was
aimed at the wrong names.

`transpilePackages` matches the **named package only, not its dependencies** — it builds a
path matcher against `node_modules/<name>/`. Verified against `pnpm-lock.yaml`:

- `@solana/kit` is a **barrel over 22 separately-published siblings**, one of which is
  `@solana/offchain-messages` — the package #209 identifies as publishing the exact
  failing `(u ||= {})` snippet. Naming `@solana/kit` covers **1 of 45** `@solana/*`
  packages and specifically not that one.
- `viem`'s dependencies are `@noble/curves`, `@noble/hashes`, `@scure/bip32`,
  `@scure/bip39`, `abitype`, `isows`, `ox`, `ws`. Naming `viem` covers none of them —
  including `@noble/hashes`, which is one of the three remaining sources.

The `browserslist` row in that table is misleading for a different reason: browserslist has
[...]

PR #251, opened 2026-08-21
feat(analytics): instrument map mount so a freeze on entry is visible

### LINKED ISSUE #196: MiniPay freezes and force-closes on entering Mondeto — player cannot buy at all

GigaHierz:
Branched off `main` (8ac29bd), independent of other open PRs.

## What & why

#196 reports MiniPay hanging for minutes on entry and force-closing, and today that population is structurally absent from analytics: a client that freezes on the map's full-grid read never fires an event. This adds a mount lifecycle to `usePixelMap` — the hook that owns the `getPixelBatch` read every map entry goes through — so a freeze shows up as a `map_mount_started` with no `map_mount_completed`.

Events (schema documented in `apps/web/src/lib/analytics.ts`, the existing single source of truth):

| event | fires | properties |
|---|---|---|
| `map_mount_started` | first commit of the map screen, before the `load()` effect issues the grid read | `mapId`, `trigger: entry\|switch` |
| `map_mount_completed` | first commit in which a full grid for the current map is ready and painted | `mapId`, `trigger`, `elapsedMs`, `attempts` |
| `map_mount_failed` | the read failed every rung of the existing retry ladder (1.5 s, 4 s, 8 s) | `mapId`, `trigger`, `elapsedMs`, `attempts`, `category`, `detail` |

Freeze rate = `1 − (completed + failed) / started`; the `elapsedMs` distribution on `completed` shows how close the slow tail runs to the host's not-responding watchdog. Every mount ends in exactly one of completed / failed / nothing, so the three nest.

`isMiniPay` is not passed explicitly: `registerPlatform()` (`apps/web/src/components/posthog-provider.tsx:84-87`) registers it as a PostHog super-property, so it is attached to all three events the same way it is to `pixel_buy_*`.

Files: `apps/web/src/hooks/usePixelMap.ts` (lifecycle), `apps/web/src/lib/analytics.ts` (schema comment), `apps/web/src/__tests__/hooks/usePixelMap.test.ts` (new).

## What this does NOT do

- **It does not fix #196; it measures it.** No change to when or how the grid is read, retried, or painted — the diff adds ref bookkeeping and three `track` calls. `git diff` shows zero changes to any `setLoadState` / `setVersion` call or effect condition that existed before.
- **No cache flag.** The issue suggested carrying "whether the read was served from cache". There is no cache in front of this read: `usePixelMap` → `fetchAllPixelsFromContract` → `readContract` directly, and `grep -rni cache src/hooks src/lib` finds sessionStorage caches only for owned-maps, analytics, snapshots and should-open-next-map. A flag that is always `false` would be noise, so it is omitted and the schema comment says why. One line to add if a cache ever lands.
- **Does not create the PostHog insight.** The freeze-rate number is a formula over three events; the insight itself has to be built in PostHog (acceptance box 4). Not verifiable from the repo.
- **Does not verify segmentation inside PostHog** (acceptance box 3). The super-property mechanism is the same one `pixel_buy_*` already segments on, but the first real events need a look on the preview/production project after deploy.
- Two pre-existing defects seen while reading this path are left alone and would each get their own issue (proposing, not filing): (a) a read that is superseded by a map switch still writes its stale pixels and flips `loadState` to `ready` (`usePixelMap.ts`, `load()` success path) — this PR only stops it counting as a completion; (b) returning continent players pay a wasted map-0 read on every entry because the stored map id is restored one effect after first render (`useMaps.ts` `CurrentMapProvider`).

## Judgement calls (all reversible, each a few lines)

1. **Where `started` fires.** A `useEffect([])` in the hook, declared before the `load()` effect, so it runs in the same commit and earlier. It cannot fire before the land-mask "decode": the masks are pre-expanded `Uint8Array` literals (`src/data/masks/*.ts`) evaluated at module import, i.e. during bundle evaluation, before any React code. The read is the heavy runtime step and `started` precedes it.
2. **Where `completed` fires, and why "painted" holds.** A passive effect on `[loadState, version]` in the hook, whose owner is the map screen. React runs descendants' passive effects first, and `InnerCanvas`'s `drawPixels` effect (`WorldCanvas.tsx`) depends on the same `version`/`pixelData` that flip in that commit — so the grid is drawn before the event fires. This is reasoned from React's effect order, not measured in a browser; if a reviewer wants it measured, a `requestAnimationFrame` after the effect would move the timestamp to after the next frame (one line).
3. **Re-fire on map switch: yes, tagged `trigger: 'switch'`.** Same heaviest read, same freeze mechanism, one filter to isolate the cold-open number. Dropping the re-fire is deleting the `else`-less branch in the contract-change effect.
4. **A switch while a mount is pending does not restart it.** `CurrentMapProvider` restores the stored map id in an ancestor effect *after* the hook's first effects, so every returning continent player mounts on map 0 and moves to map N a tick later. Restarting would make every such entry a start-with-no-completion and corrupt the freeze rate — the metric this exists for. Instead the pending mount carries on and `completed.mapId` is the map that actually painted; `started.mapId` may be the pre-restore id. A generation counter makes sure the superseded map-0 read can't close the mount early. `attempts` resets on the carry-over so it means "reads of the map that painted".
5. **`failed` is terminal only.** A single rejected read is not a failed mount while the bounded retry ladder is running; a read that fails once and recovers completes with `attempts: 2`. Firing on the first rejection would count every transient Forno blip as a failure and understate the very freezes the pair exists to surface.
6. **Error category reuses `categorizeBuyError`** (`lib/buyErrors.ts`) rather than a second classifier — its `timeout` / `rpc` / `unknown` rules are the ones a read can hit and are what `pixel_buy_failed` already segments on. `detail` is the first line of the error only (viem puts the URL, with any RPC key in the query string, on later lines), truncated to 100 chars; pinned by a test asserting the `https://` never appears.

## Verification

Run on the commit in this PR, worktree at `apps/web`:

```
$ pnpm --filter web test:coverage
 Test Files  59 passed (59)
      Tests  540 passed (540)
Statements   : 37.87% ( 1708/4509 )   # main: 35.34%
Branches     : 32.25% ( 827/2564 )    # main: 31.22%
Functions    : 38.16% ( 321/841 )     # main: 35.44%
Lines        : 38.48% ( 1509/3921 )   # main: 35.88%

$ pnpm --filter web type-check
> tsc --noEmit
(exit 0)
```

Pre-fix red: the new file against the unmodified hook fails 7/7 (`expected [] to deeply equal [{ mapId: 0, trigger: 'entry' }]`).

Tests (`src/__tests__/hooks/usePixelMap.test.ts`, through the hook with `track` and `useReadClient` doubled; the read double routes by `functionName` and contract address and returns one stable client like the real hook):

- started fires before the read resolves; completed after, in that order, `elapsedMs: 1234` after a 1234 ms fake-clock advance, `mapId: 0`
- a read held open for 60 s produces `started` only — no completed, no failed (acceptance box 2)
- completed fires once across re-renders, a `refresh()` and a 30 s poll tick (paired with the positive case above so the absence can fail)
- a read that fails all four attempts fires `failed` once, after the last rung and not before, with `category: 'rpc'`, `attempts: 4`, `elapsedMs: 13500`, `detail` without the URL; never completed; nothing re-fires 60 s later
- a read that fails once then succeeds completes with `attempts: 2`, never fails
- a switch after completion starts a `switch` mount and completes against map 1
- a switch before completion keeps one `entry` mount; the late map-0 read does not complete it, the map-1 read does

Mutation tally (one mutation at a time on `usePixelMap.ts`, then restored and byte-compared):

| mutation | red |
|---|---|
| M1 drop the `map_mount_completed` emission | 5 / 7 |
| M2 fire `started` only after the read (order swapped) | 4 / 7 |
| M3 a superseded read marks readiness | 1 / 7 |
| M4 `failed` fires on the first error instead of the last retry | 2 / 7 |
| M5 `completed` ignores the generation check | 2 / 7 |
| M6 `completed` never closes the mount | 2 / 7 |

Lint: `pnpm --filter web lint` fails in my worktree before reaching any file — ESLint's legacy cascade walks up past the worktree into a parent directory's `.eslintrc.json` and cannot resolve `next/core-web-vitals` from there. Control run on the stashed clean tree (0 changed files) fails identically, so it is the local layout, not this diff. Same config run in isolation:

```
$ npx eslint --no-eslintrc -c .eslintrc.json --ext .ts,.tsx src/hooks/usePixelMap.ts src/lib/analytics.ts src/__tests__/hooks/usePixelMap.test.ts
(exit 0, no output)
$ npx eslint --no-eslintrc -c .eslintrc.json --ext .ts,.tsx src
✖ 22 problems (0 errors, 22 warnings)   # all pre-existing, none in the changed files
```

CI's `next lint` on a normal checkout is the authoritative run.

Not verified automatically: the events arriving in PostHog with `isMiniPay` attached, and the "painted before `completed`" ordering in a real browser. Both need one look at the preview deployment's PostHog project.

## Merge order

Independent; no shared files with other open PRs that I know of. Whoever touches `usePixelMap.ts` next rebases.

Refs #226 — closes acceptance boxes 1 and 2 by test; box 3 is wired (super-property) but not yet observed in PostHog; box 4 needs the insight built.
Refs #196, #225.


csacanam (changes requested):
Reviewed at head 315b908, re-fetched and compared against headRefOid immediately before posting.

REQUEST-CHANGES — two findings. The lifecycle modelling is the good part of this PR and I would not touch it; both findings are about the data the events carry.

Calling this "instrumentation" undersells it. It adds a generation counter and a mount state machine to a hook that renders the main screen, and the race reasoning in the comments is the substance. I reviewed it as product code, which is what it is.

FINDING 1 — categorizeBuyError(String(err)) is the narrow-haystack defect that #246 is open to fix, in a sibling path.

  category: categorizeBuyError(String(err)),

String(err) gives Name: message for the top-level error only. It does not reach cause.details or cause.shortMessage — which is precisely the blind spot #246 documents: for viem's HttpRequestError the classifiable text lives on the cause (shortMessage "HTTP request failed.", details "too many requests"), and a classifier handed only the outer string files it as unknown.

So map_mount_failed will under-report rpc for exactly the failure this event exists to catch. A rate-limited grid read on entry — the freeze you are instrumenting — is the most likely thing to land here, and it is the case most likely to be miscategorised.

This is the same shape as the defect #246 names: a fix landing in one path while a sibling path that serves users keeps the identical defect. Worth fixing here rather than filing, because the two PRs are open at the same time and both are yours.

Shape of the fix: export collectErrorText from #246 (or lift it to lib/buyErrors.ts next to the classifier, which is arguably where it belonged all along) and call categorizeBuyError(`${String(err)} ${collectErrorText(err)}`). If you would rather not couple the two PRs, land #246 first and rebase this on it. Either is fine; leaving it is not.

Second-order point, lower confidence: the buy classifier's first six rules are not_land, token_not_accepted, slippage, deadline_expired and insufficient_funds — buy-path revert categories that cannot occur on a read-only grid fetch. Reusing the classifier is reasonable for the transport half (rpc, timeout, gas_estimate) but it means map_mount_failed can emit a category that is meaningless for a read. Not blocking, and I would not build a second classifier for it, but a line in the schema saying which categories are actually reachable here would save whoever queries this a confused hour.

FINDING 2 — map_mount_started and map_mount_completed can report different mapIds for the same mount, and nothing says so.

The entry event fires from an effect with an empty dependency array, so resolvedMapId is captured at first render:

  useEffect(() => {
    if (mountRef.current) return
    mountRef.current = { startedAt: Date.now(), trigger: 'entry', attempts: 0 }
    track('map_mount_started', { mapId: resolvedMapId, trigger: 'entry' })
  }, [])

The completed event reads resolvedMapId at completion time. Your own comment describes the case where these differ — the stored-map restore that runs one effect after first render — and correctly decides the pending mount should carry on rather than restart, so the restore race cannot produce a start with no completion. That decision is right.

But the consequence lands in the data: started says mapId 0 and completed says mapId 3 for one mount. Anyone analysing this will join started to completed on mapId, and the entry mounts affected by the restore — which is the default path for any returning player — will look like a start that never completed plus a completion that never started. That inverts the exact signal the PR is built to produce.

Shape of the fix, small: put mapId in PendingMount alongside startedAt and trigger, and emit that stored value on completed and failed. Both ends then describe the same mount. If you would rather report the map that actually painted, add it as a separate field (paintedMapId) rather than overloading mapId — but do not leave the two ends disagreeing silently.

What I verified as sound, so you know what not to touch:

- The generation counter is the right mechanism and the invariant is stated correctly: a load() captures the generation it started for, so a read landing after the map moved on still writes its stale pixels exactly as before but is never taken as this mount's completion. Behaviour preserved, measurement corrected.
- Ref-guarding the entry effect against StrictMode's double invoke is necessary and easy to forget.
- Keying the completed effect on version as well as loadState is not belt-and-braces: a superseded read can park loadState on 'ready' before the real one lands, and without version the effect would not re-run. That is the subtle one and you got it.
- Treating failed as terminal-only — after the retry ladder is exhausted, not on the first rejected read — is the correct definition. A single failure while retries are still running is not a failed mount, and if a retry paints the grid the mount completes slowly instead. That is what makes elapsedMs meaningful.
- Resetting attempts rather than restarting the mount on a switch keeps attempts as "reads of the map that painted," which matches what the field name promises.
- I checked whether retryCountRef leaks across a map switch, since a stale count would fire failed with attempts: 0 on the first error of a new map. It does not — line 86 resets it in the same effect. Not an issue.
- Placing the completed effect on the hook's owner and relying on React running descendants' effects first, so the canvas has already painted when it fires, is correct and is the kind of thing worth the three lines of comment you gave it.

Nothing here changes what a player sees, and no product code path is altered beyond the emissions — I confirmed the load/retry/refresh behaviour is byte-for-byte the same apart from the ref bookkeeping.

Once the two findings are addressed this is a straightforward approve. It is one round of targeted fixes, not a rethink.

GigaHierz:
Both findings fixed. Pushed `1546128`. **Base retargeted to `GigaHierz/215-untested-buy-failure-emissions` (#246)** — this now stacks on it, for the reason in finding 1. Merge #246 first and this retargets cleanly.

## FINDING 1 — `categorizeBuyError(String(err))` in a sibling path · **FIXED**

Confirmed rather than taken on reading. The existing fixture hid it: `RATE_LIMIT_ERROR` is a **flat** `Error` with everything on the top-level message, so `String(err)` already classified it `rpc` and the suite was green. The real viem shape puts the classifiable fields on the cause and leaves the outer generic. With that shape:

```
× classifies a rate limit whose text is only on the cause (not `unknown`)
```

Now classifies against `collectErrorText` — **lifted to `lib/buyErrors.ts` in #246**, taking your "arguably where it belonged all along" option rather than exporting it from the hook. Hence the stack.

**A third defect fell out of writing that test, and it is the more interesting one:**

```ts
rejectLatest: async () => {
  calls[calls.length - 1].reject(RATE_LIMIT_ERROR)   // ignores its argument
```

The helper took no error and always rejected with the flat one. My test passed the nested shape and the harness threw it away — the test asserted nothing and passed. That is checklist #10 exactly, and it is why I mutation-tested rather than trusting green. `rejectLatest` now honours what it is handed.

## FINDING 2 — started and completed report different mapIds · **FIXED**, taking your alternative

`PendingMount` carries the map the mount started on, and both ends emit it. **I took your `paintedMapId` suggestion rather than just storing `mapId`**, because the existing switch test showed why it matters:

```
- mapId: 1     ← what completed used to say
+ mapId: 0     ← what started said all along
```

That assertion was **encoding the disagreement**. Storing the mount's id alone would have made the events joinable but silently discarded which map actually rendered — a real fact that test was (accidentally) protecting. So both are kept: `mapId` is the mount, `paintedMapId` is what painted.

**`paintedMapId` is emitted only when the two differ.** Adding it unconditionally churned three unrelated event-shape assertions for no information — and making its presence conditional is what makes it meaningful: it appears exactly when a restore moved the map mid-mount, which is the thing worth counting.

**I edited an existing test.** Only that one, only the `mapId`, plus the new `paintedMapId`, with a comment saying what it used to assert and why. Flagging it because a fix that quietly rewrites an assertion is a new unreviewed change.

## Mutation count — 2 mutations, 3 red

| mutation | red |
|---|---|
| revert the haystack widening to `String(err)` | **1** — the nested rate-limit test |
| revert the mapId fix (read the current map at completion) | **2** — the new join test **and** the existing switch test |

That second row is the useful one: the existing test now fails in the *right* direction, so the behaviour is pinned from both sides.

## On your second-order point — taken, not built

You were right that the buy classifier's first six rules (`not_land`, `token_not_accepted`, `slippage`, `deadline_expired`, `insufficient_funds`) are buy-path reverts that cannot occur on a read-only grid fetch, and that a second classifier would be the wrong cure. I have not built one. #246 now pins the full rule table, so the reachable set is at least legible from the tests — but the schema line you asked for is a good idea and I would rather write it once the CAMPAIGN-era events settle than guess now.

## What I did not touch

Everything in your "verified as sound" list — the generation counter, the StrictMode ref guard, keying the completed effect on `version`, terminal-only `failed`, the attempts reset on switch, the effect-ordering comment. No lifecycle behaviour changed; only the payloads.

```
$ corepack pnpm run test        # on 1546128
  Test Files  64 passed (64)
  Tests  637 passed (637)
$ corepack pnpm run lint         # 0 errors
$ corepack pnpm run type-check   # 0 errors
```

Ready for re-review — after #246.

csacanam (dismissed):
APPROVE. Re-reviewed at head 1546128. Both findings fixed, and both fixes taught something — which is what a resolution round is for.

FINDING 1 (narrow haystack in a sibling path) — FIXED. categorizeBuyError now gets `${String(err)} ${collectErrorText(err)}`, importing from lib/buyErrors.ts where #246 lifted it. Right call on the placement question: the lib is where it belongs, and the retarget of this PR's base onto #246's branch handles the ordering mechanically — I verified the stack (this PR's diff is only its own 3 files, and #246's head is its merge-base). Merge order: #246 → this, and this retargets on #246's merge.

What the fix taught is the part I want to underline: my case-3 instinct was right for a reason neither of us stated — your harness's rejectLatest ignored the error it was handed and always threw the flat fixture, so the nested-error test passed while asserting nothing. That is test-double honesty (checklist #10) caught by mutating your own fix, and the confession is now a comment on the fake itself, which is where it belongs. The NESTED_RATE_LIMIT_ERROR fixture — generic outer message, every classifiable field on the cause — is also the truest #215-shaped fixture either PR has; see my note on #246 suggesting it be borrowed there.

The URL-never-travels control on the new test (detail must not contain https://) closes the loop I did not ask for: widening what the classifier reads without widening what we send. Good.

FINDING 2 (started/completed disagree on mapId) — FIXED, and your alternative is better than my suggestion.
I proposed storing the mount's mapId and emitting it on both ends. You took the paintedMapId variant and the reasoning is the decisive part: the existing switch test was asserting mapId: 1 on completed while started said 0 — the suite was encoding the bug — and storing only the mount's id would have thrown away which map actually painted. Keeping both, with paintedMapId emitted only when they differ, means: joins on mapId are clean, the restore race is visible instead of silent, and the common case carries no extra field. Verified in the diff: PendingMount carries mapId, started/completed/failed all emit mount.mapId, paintedMapId appears only on divergence.

Editing the one existing assertion that encoded the bug, and flagging the edit rather than slipping it through, is exactly how a changed assertion should land.

One non-blocking note carried over from my first review, unchanged: the schema comment in analytics.ts could still name which categories are actually reachable on a read path (the revert categories can't occur on a grid fetch). Not worth a round; fold it into any future touch of that file.

GigaHierz:
**My mistake, and it cost your approval on this PR — apologies.**

I merged #246 with `--delete-branch` while this PR was still based on that branch. Deleting the base branch of an open PR closes it, so GitHub auto-closed this at 11:25:18Z and auto-retargeted it to `main`. Reopening it lost the approval.

Nothing was at risk — the head branch survived — but the correct order was to retarget this to `main` *first*, then merge #246. Recording it because it is the second stack-mechanics trap in this queue and both are cheap to avoid:

- **Merging the base of a stack with `--delete-branch` closes the child.** Retarget the child first.
- **`gh pr update-branch` preserves approvals; a force-pushed rebase dismisses them.** That is why #202/#206/#246 kept yours through their updates while #73 and #248 lost theirs earlier — same ruleset, different push shape.

**Rebased onto `main`** with `git rebase --onto origin/main 04006d8`, dropping #246's two commits now squashed onto `main`. This is back to its own two commits and the diff is unchanged from what you approved.

```
$ corepack pnpm run test        # on c83e078
  Test Files  64 passed (64)
  Tests  638 passed (638)
$ corepack pnpm run lint         # 0 errors
$ corepack pnpm run type-check   # 0 errors
```

Needs a re-approve. Content is identical to the head you already reviewed — the only change is that #246's commits are no longer underneath it.

csacanam (approved):
RE-APPROVE after the accidental close/reopen. Verified the identical-content claim with git range-diff: both of this PR's own commits are patch-identical to the head I approved (50127c5 = 0773918, 1546128 = c83e078); everything else that moved is #246's squash and the dep bumps now on main, absorbed by the rebase. ci / ci green on head c83e078. No fault taken on the auto-close — the mechanics finding you filed from it (update-branch preserves approvals) is worth more than the clicks it cost.

Issue #247, opened 2026-08-19
chore(minipay): extract the on-screen error overlay from #204 — errors are unreadable in MiniPay's WebView


GigaHierz:
## Problem

When the app dies inside MiniPay, we cannot read the error. MiniPay renders miniapps in the device's **Android System WebView, which is not remotely inspectable** — with MiniPay running, the only devtools socket on the device is Chrome's. So Next's "a client-side exception has occurred (see the browser console)" is a dead end on exactly the devices that fail: there is no console to see.

This has already cost us real time. #196 sat on support reports as its entire evidence base for weeks, and the root cause was only found once someone painted the error onto the screen of the failing handset:

```
ERROR: Uncaught SyntaxError: Unexpected token '='
  at /_next/static/chunks/286-cd92f5e4412c85c4.js:1
```

That output is what identified logical assignment operators (`||=` / `??=`, Chrome 85+) as the parse failure now tracked in #225. Without it, the investigation had a blank blue screen and nothing else.

## What exists, and where

Draft PR #204 (`csacanam/minipay-old-webview-blank-map`, head `b243bedb21872a0e962839cd95b86ffa51c46c50`) contains the working overlay:

- **`apps/web/src/lib/debugErrorOverlay.ts`** — installs `onerror` + `unhandledrejection` handlers that paint message, source and line into a fixed banner. Opt-in via `?debug=1`; never runs for normal traffic.
- **`apps/web/src/app/layout.tsx`** — inlines it as a `beforeInteractive` `<Script>`, so it is armed *before any chunk evaluates* and catches failures during bundle execution. That ordering is the whole point: a parse error in a chunk happens before any React error boundary exists.
- Written **ES5-only**, deliberately, because it has to run on the engine that just failed to parse everything else.

#225 says this half of #204 "is independently useful and should be extracted separately". This is that issue — filing it so the work outlives the draft branch rather than living in one sentence.

## Why extract it rather than merge #204

#204 is a diagnosis, not a fix, and is draft on purpose. Its other half — `legacyShim.ts`, an ES5 shim for `Object.hasOwn` and `String.prototype.replaceAll` — is explicitly superseded: #225 argues a hand-written two-function shim is the wrong answer and `core-js@3` driven by browserslist is the durable one. Merging #204 whole would ship the superseded half; leaving it draft loses the half that works.

## Scope

[...]

PR #252, opened 2026-08-21
chore(minipay): paint uncaught errors on screen behind ?debug=1 — MiniPay's WebView has no console

### LINKED ISSUE #196: MiniPay freezes and force-closes on entering Mondeto — player cannot buy at all

GigaHierz:
## What & why

When the app dies inside MiniPay there is nothing to read: MiniPay renders mini-apps in the device's Android System WebView, which is not remotely inspectable, so Next's "a client-side exception has occurred (see the browser console)" is a dead end on exactly the devices that fail. The parse failure behind #196 (`||=` / `??=` in dependency chunks, Chrome 85+, now tracked as the build floor in #225) was only identified once the error was painted onto the handset's screen — the diagnosis on #204.

This PR lifts that tool out of draft #204 (head `b243bedb21872a0e962839cd95b86ffa51c46c50`) onto a branch off `main`, without `legacyShim.ts` or its tests (superseded by #225):

- `apps/web/src/lib/debugErrorOverlay.ts` — an ES5-only IIFE that, when the URL carries exactly `debug=1`, installs `window.onerror` + an `unhandledrejection` listener and appends `ERROR: <message> at <source>:<line>` / `UNHANDLED REJECTION: <stack|message>` lines to a fixed banner, plus one `UA:` and one `HAS:` (`Object.hasOwn` / `String.prototype.replaceAll` presence) line on `DOMContentLoaded`. Display-only; nothing is transmitted.
- `apps/web/src/app/layout.tsx` — inlines it in `<head>` as a **raw `<script>`** (see judgement calls: this deliberately differs from #204 and from the issue text, on measurement).
- `docs/MINIPAY_BUILD_PLAYBOOK.md` — new "Debugging in MiniPay's WebView" section next to the MiniPay gotchas: opt-in, what it shows, what it cannot do.
- Tests: `__tests__/lib/debugErrorOverlay.test.ts` (ES5 guard + behaviour through the seam, in a fresh jsdom window per case), `__tests__/app/layout.test.ts` (mount mechanism), `__tests__/jsdom.d.ts` (ambient types for the slice of `jsdom` the test uses).

Refs #247 — closes acceptance boxes 1 (extracted without the shim), 2 (inert without the flag / paints on `onerror` *and* `unhandledrejection` with it) and 3 (arms before chunk evaluation — confirmed by measurement, which changed the mechanism). Box 4 (real device at the Chrome 80 floor) is **not** met by this diff; see "What this does NOT do".
Refs #196, #225. Draft PR #204 can be closed once this merges: its overlay half lands here and its shim half is superseded by #225.

## What this does NOT do

- **Does not fix #196 or #225.** It cannot make an old engine parse new syntax; it makes the failure readable. Zero runtime behaviour changes for requests without `debug=1` — the script's first statement returns.
- **Not verified on a real Chrome 80 WebView.** Everything below ran in headless Chrome 151 and jsdom. The ES5 property is pinned by a token guard, not by running the payload on Chrome 80. Box 4 of #247 stays open until someone loads `?debug=1` on a floor device and reads the banner.
- **Not a parser-grade ES5 check.** `acorn` is not a declared dependency of `apps/web`, so the guard is a pinned list of 14 post-ES5 token patterns (arrow, `const`/`let`, template literal, `?.`, `??`, spread, `class`, `for…of`, `async`/`await`, generators, `||=`/`&&=`, `**`, default params). It does not detect object-literal shorthand methods — in minified code `{show(` is indistinguishable from a block starting with a call. Extend the list, never relax it.
- **Does not redact.** What it paints is whatever the engine hands `onerror` / a rejection's `stack`/`message`. A rejected viem call can carry the connected account address in its message; that is the opted-in user's own address on their own screen, and nothing leaves the device. No wallet- or address-shaped data is *collected* by this code.
- **Does not gate by environment.** `?debug=1` works in production on purpose — the whole value is reading a player's handset in the field; `/dev/*` gating intentionally does not apply. Anyone can append the param and see their own uncaught errors. That is the decision #247 asks to have stated; say the word and a `VERCEL_ENV` gate is a one-liner.

## Judgement calls

1. **Raw `<script>`, not `<Script strategy="beforeInteractive">` — on measurement, not taste.** #204 and #247 both say `beforeInteractive` is "the load-bearing part". Building and serving it showed otherwise. In the App Router, `next/script` emits no script for `beforeInteractive`; it serialises the body into `(self.__next_s=self.__next_s||[]).push([0,{children:"…",id:"debug-error-overlay"}])` (`node_modules/next/dist/client/script.js:274-296`), and that queue is only drained by `appBootstrap()` in the `main-app` chunk right before hydration (`node_modules/next/dist/client/app-bootstrap.js:22-61`). So the handlers are installed only after the webpack runtime, the React chunk and `main-app` have all parsed and executed — and `main-app` requires React *before* it reaches `appBootstrap`. Measured with a local proxy replacing one chunk's body with a parse error and headless Chrome dumping the DOM at `/faq?debug=1`:

   | poisoned chunk | `<Script beforeInteractive>` banner | raw `<script>` banner |
   |---|---|---|
   | none (baseline) | `UA:` line only (the `__next_s` drain ran) | `UA:` + `HAS:` |
   | React framework `21d6c11b` | **NONE** — overlay never ran | `ERROR: Uncaught SyntaxError: Unexpected token '}' at …/21d6c11b-….js:1` |
   | `main-app` | **NONE** | `ERROR: … main-app-….js:1` |
   | `webpack` runtime | **NONE** | `ERROR: … webpack-….js:1` |

   The raw script sits at byte 1825 of the served `/faq` document, inside `<head>`, with 15 `<script async src>` tags before it in document order and **zero** synchronous external scripts — so per the HTML parsing model it executes before any chunk can, and depends on nothing else parsing. On the #196 handset the `beforeInteractive` version did work, because the chunks that failed there (`286`, `d0466ee9`) are page-level and arrive after the bootstrap path; it would have been blind to a failure in the bootstrap path itself. One-line revert if you disagree: swap back to `<Script strategy="beforeInteractive">` — the layout test (`M5'` below) goes red so the choice has to be made consciously.
2. **Gate tightened from `indexOf("debug=1")` to `/[?&]debug=1(&|$)/`.** #204's substring check also armed on `?debug=10` and `?xdebug=1`. Exact-parameter match is what the issue and the docs promise; pinned by the five inert cases in the test. Trivial to loosen if wanted.
3. **Ambient `jsdom` types instead of `@types/jsdom`.** `jsdom` is already a devDependency (vitest's environment) but ships no types. Declaring the two members the test uses keeps the lockfile untouched (lockfile PRs have a history here). If `@types/jsdom` is preferred, delete `__tests__/jsdom.d.ts` in the same change — the two would collide.
4. **Kept `window.onerror` assignment rather than `addEventListener('error')`.** It is what ran on the device; PostHog's exception autocapture wraps and chains a pre-existing `onerror`, so it survives. Not measured against PostHog here — flagging, not asserting.
5. **Docs location.** The note went into `docs/MINIPAY_BUILD_PLAYBOOK.md` (the only MiniPay doc with operational content) rather than `apps/web/CLAUDE.md`, which is about logging conventions.

## Verification

All on the final tree, macOS, Node 20.19.4, pnpm 8.10.0.

```
$ pnpm --filter web test
 Test Files  60 passed (60)
      Tests  564 passed (564)        # 533 on main + 31 new (28 overlay, 3 layout; no existing test touched)

$ pnpm --filter web test:coverage ; echo exit=$?
      Tests  564 passed (564)
Statements 35.35% · Branches 31.22% · Functions 35.44% · Lines 35.9%   (floors 28/25/28/29 held)
exit=0

$ pnpm --filter web type-check ; echo exit=$?
exit=0   (0 × "error TS")

$ npx eslint --no-eslintrc -c .eslintrc.json --ext .ts,.tsx src    # in apps/web
✖ 22 problems (0 errors, 22 warnings)   # identical set to main; the 5 touched files add none
                                        # beyond the 2 pre-existing no-page-custom-font warnings
                                        # on untouched <link> lines in layout.tsx

$ pnpm --filter web build ; echo exit=$?
ƒ  (Dynamic)  server-rendered on demand
exit=0
```

Note on lint: plain `next lint` from this worktree fails to *load* config because `apps/web/.eslintrc.json` has no `root: true` and ESLint walks up into an unrelated parent checkout on this machine; CI checks out at repo root and is unaffected. The pinned-config invocation above is the same rule set.

**TDD / red-first.** Both test files were written before the implementation and run red on `main`'s tree: the overlay file could not be imported (whole file fails) and both layout assertions failed.

**Mutation count — 7 mutations, each turning at least one test red, sources restored byte-identical after each:**

| # | mutation | red |
|---|---|---|
| M1 | delete the `debug=1` gate line | 5 (all "installs nothing and paints nothing" cases) |
| M2 | add `const x=1;` to the payload | 1 (`contains no const declaration`) |
| M3 | make the `onerror` handler an arrow function | 1 (`contains no arrow function`) |
| M4 | delete the `unhandledrejection` listener | 2 (both rejection tests) |
| M5' | mount via `<Script strategy="beforeInteractive">` (the #204 shape) | 2 |
| M7 | add `async` to the raw `<script>` | 1 |
| M6' | `main`'s layout (no overlay) | 3 |

**Seam checks.** Overlay tests evaluate the *string* in a fresh `JSDOM` window per case via `window.eval`, then dispatch real `ErrorEvent`s and `unhandledrejection` events at it — absence cases assert `window.onerror === null` **and** an empty `<body>` after both events and after `DOMContentLoaded`, paired with presence controls. The ES5 guard has two controls: the payload is >200 chars and parses under `new Function`, and a deliberately poisoned copy trips the guard. Rendered-HTML and poisoned-chunk measurements are in judgement call 1.

**Not verified:** a real Chrome 80 Android WebView (box 4 of #247). Please load `https://<preview>/?debug=1` on a floor device before closing the issue; the expected banner on a healthy engine is two lines (`UA:` and `HAS:`).

## Merge order

Branched off `main` (8ac29bd), independent of every other open PR; touches `layout.tsx`, which no open PR other than draft #204 edits. No lockfile change. After merge: close #204 with a pointer here and #225 (its two halves are accounted for); #247 stays open for box 4 unless the maintainer decides the device check belongs to #225's rollout instead.


csacanam (approved):
Reviewed at head c6031b9, re-fetched and compared against headRefOid immediately before posting.

APPROVE. Zero findings. Answering the decision you asked for, plus one follow-through that is mine to do, not yours.

THE DECISION YOU ASKED FOR — keep ?debug=1 working in production, ungated.

You are right that this is the call #247 wants stated rather than assumed, so here it is on the record.

Keep it. The entire value is reading a player's handset in the field, and that player is on production. A VERCEL_ENV gate would leave us exactly where #196 left us: a dead end on the only devices that fail. The /dev/* gating convention is about not shipping internal tooling to players; this is a diagnostic a player opts into on their own device.

The exposure is bounded in the way that matters and you characterised it accurately: opt-in via an exact parameter, display-only, nothing transmitted, and what it paints is the user's own error on the user's own screen. Your note that a rejected viem call can carry the connected account address is the right disclosure — and it is their address, visible to them, on a screen they already control. Nothing is collected.

So: no gate. The docs section is what makes this a supported tool rather than a leftover, and it is in the right file.

JUDGEMENT CALL 1 — the raw <script>. Keep it, and this is the best-argued decision in any PR I have read this week.

You contradicted both #204 and the issue text, on measurement, and showed your work. In the App Router, beforeInteractive emits no script — it serialises into the self.__next_s queue drained by appBootstrap in main-app, so the handlers install only after the webpack runtime, React and main-app have all parsed. Your poisoned-chunk table is the proof:

  poisoned chunk         beforeInteractive        raw <script>
  React framework        NONE                     ERROR: … 21d6c11b-….js:1
  main-app               NONE                     ERROR: … main-app-….js:1
  webpack runtime        NONE                     ERROR: … webpack-….js:1

An overlay that goes blind exactly when the bootstrap path is what broke is worse than no overlay, because it looks like the app is fine. Explaining why the #204 version appeared to work on the #196 handset — the failing chunks there were page-level and arrived after bootstrap — is what turns this from "I changed it" into "here is why the old evidence was misleading."

I checked the one thing that could invalidate the raw-script approach and it is clear: there is no Content-Security-Policy anywhere in the app or vercel.json, so dangerouslySetInnerHTML in <head> executes. Worth knowing that if a CSP is ever added, script-src will need a hash or nonce for this tag or the overlay silently stops arming — the failure mode being silence, that is worth a line in the playbook section at some point. Not blocking, not this PR.

The other judgement calls are all right: tightening the gate from indexOf("debug=1") to /[?&]debug=1(&|$)/ closes ?debug=10 and ?xdebug=1; ambient jsdom types keep the lockfile untouched, which given the five-way lockfile conflict in this repo right now is worth more than usual; and putting the docs in MINIPAY_BUILD_PLAYBOOK.md rather than apps/web/CLAUDE.md is correct since the latter is about logging conventions.

Judgement call 4 is the one soft spot and you labelled it correctly as flagging rather than asserting: window.onerror survives PostHog's autocapture because PostHog wraps and chains a pre-existing handler, but you did not measure it. I agree with the choice — it is what ran on the device — and I am not asking you to measure it now. If the banner ever comes up empty on a device where PostHog is live, that is the first thing to check.

Verified:

  pnpm --filter web test    # 60 files, 564 tests

Mutation table is honest and M5' is the one that earns its place: mounting via <Script strategy="beforeInteractive"> — the #204 shape — turns 2 red. That means the mechanism decision cannot be reverted by accident, which is exactly what you want for a choice that took a proxy and a poisoned chunk to establish.

The seam work is right: evaluating the payload string in a fresh JSDOM window per case and dispatching real ErrorEvents at it, with absence cases asserting window.onerror === null AND an empty body after both events and DOMContentLoaded, paired with presence controls. And the ES5 guard has two controls of its own — the payload is >200 chars and parses under new Function, and a poisoned copy trips the guard. Without the first control, a guard that silently received an empty string would pass.

You are appropriately clear about what the token-list guard is not: not parser-grade, blind to object-literal shorthand methods in minified code, extend the list never relax it. Since the payload is hand-written rather than minified, that limitation does not bite here, but writing it down is what stops someone reusing the guard somewhere it would.

ON #204 — the follow-through is mine and I will do it.

You lifted this out of my draft and pinned its head SHA, b243bedb21872a0e962839cd95b86ffa51c46c50. I checked: that is #204's actual head. Citing the exact commit rather than "from the draft PR" is what makes the provenance checkable after the branch is gone, and it is the right way to carry unmerged work forward.

I will close #204 once this merges, with a comment pointing here for the overlay half and at #225 for the shim half, so the closure says where the value went rather than just going quiet. You do not need to do anything about it.

Refs #247 with box 4 left open is correct, and box 4 — loading ?debug=1 on a real device at the Chrome 80 floor — is the same open box #253 is gated on. Worth resolving those two together rather than twice.

PR #270, opened 2026-09-14
docs(fees): step the Terms and FAQ resale fee copy from 5% to 7.5%

### LINKED ISSUE #265: task: update Terms & FAQ copy for feeRate step 5%→7.5%
## What needs to be done?
Player-facing copy in Terms and FAQ hardcodes the resale fee as a fixed
"5%" in three places. The team is stepping `feeRate` 5%→7.5% this week
(bet #1 of the Mondeto revenue plan). Once the on-chain value changes,
this copy will be wrong — it needs to be updated in lockstep, not after
the fact.

**Verified by reading the source** at commit `a4f8536325bfd78eafc807226a31d027b29e8bb7`
(not run/executed — this is static text, no code path to execute):

- `apps/web/src/app/terms/page.tsx:61` — "Celo Core Co. may receive a
  service fee of up to five percent (5%) of the value of each purchase
  of an owned pixel"
- `apps/web/src/app/faq/content.ts:117` — "they receive what you paid
  minus a 5% platform fee"
- `apps/web/src/app/faq/content.ts:122` — "minus the 5% fee"
- `apps/web/src/app/faq/content.ts:15-16` — existing comment already
  flags these numbers "were verified against mainnet `feeRate()` and
  `config()` ... per-deployment constructor values" — i.e. the file's
  own convention already anticipates they need re-verification when the
  value changes.

**Root cause:** `feeRate` is owner-settable on-chain via `setFeeRate()`
(`apps/contracts/src/Mondeto.sol`), capped by `MAX_FEE_RATE` = 2000 bps
(20%) — so 7.5% is within bounds and requires **no smart contract code
change**, only the admin call itself. The copy above is a hardcoded
snapshot with no mechanism tying it to the live on-chain value.

**Impact:** if the fee changes on-chain before this copy is updated,
Terms & FAQ understate what players actually pay — user-visible wrong
data about money, in a legal/compliance document.

## Suggested fix (and what it does NOT fix)
1. Coordinate with Karl on the `setFeeRate()` admin call timing/execution
   (5%→7.5%). This issue tracks the copy update, not the on-chain call
   itself.
2. Update `terms/page.tsx` §6 wording and both `faq/content.ts` fee
   mentions to 7.5%, same day as the admin call — not after.
3. Re-verify the copy against mainnet `config()` per the existing
   comment convention at `faq/content.ts:15-16`, and update that
   comment's "verified against" note with the new date.

**Does NOT fix:** the later 5%→10% step is explicitly out of scope here
— see OPEN below.

## Non-goals / dependencies
- No smart contract change (feeRate is already mutable within its cap).
- Depends on the admin call landing first — the copy PR should reference
  the tx hash of the `setFeeRate()` call it corresponds to.
- Does not cover the eventual 10% step or its elasticity monitoring —
  tracked separately (see OPEN).

## Definition of done
- [ ] `terms/page.tsx` §6 fee wording matches live `feeRate` at 7.5%
- [ ] `faq/content.ts:117` and `:122` updated to 7.5%
- [ ] `faq/content.ts:15-16` comment's "verified against mainnet" note re-confirmed with new date

**How we'd know it's fixed:** grep the three cited locations for "5%" —
zero matches once the copy PR merges.

## OPEN
- The later 5%→10% step (pending no volume drop from this step) is
  explicitly NOT in scope here — needs its own follow-up issue once
  scheduled, left to implementer/PM.

ghozzza:
**Merge gate: not before the `setFeeRate(750)` call lands.** `feeRate()` on all three live maps (world `0xA8cF…5C49`, africa `0x8e70…79b7`, asia `0x9b8D…0661`) still returns **500 bps at block 77490520** (2026-09-14 13:07Z). Merge this the same day the admin call is mined and paste the tx hash here — the copy is meant to change in lockstep, and merged early it would state a fee players are not yet paying.

Player-facing copy hardcodes the resale fee at 5% in three places — Terms §6 (`terms/page.tsx:61`), and the FAQ's `who-pays-whom` and `sniped` answers (`faq/content.ts:117`, `:122`). The rate is owner-settable on chain (`setFeeRate`, capped at `MAX_FEE_RATE` 2000 bps), and the plan steps it 500 → 750. Nothing ties the copy to the live value, so the moment the call lands, a legal document and two money answers are wrong. This changes the three strings to 7.5% ("up to seven and a half percent (7.5%)" in the Terms), updates the two source comments that repeat the number (`terms/page.tsx:11`, `faq/content.ts:15-19`), and rewrites the FAQ header note so it says the fee is settable and was stepped by a tx rather than being a constructor value. The FAQ facts test now pins `7.5% ` and rejects a bare `5% `, so a partial revert of one answer fails CI.

Two things a reviewer should know. The Terms file's own header says any further edit to that text needs legal sign-off, so §6 wants counsel's OK before merge even though the change is one number — routing that is outside this PR. And the `faq/content.ts` note says the tx is "referenced by the PR that landed this line": that is this PR, and the hash goes into the first comment when it exists; the "re-confirmed" box on #265 gets ticked by a `cast call` against all three maps that day, not by this diff.

Verified: `tsc --noEmit` clean, `vitest run src/__tests__/app/faq.test.ts` 11 passed. Mutation control: reverting one FAQ answer to 5% turns the facts test red. Not verified: nothing here executes — it is static copy, so the check that matters is the on-chain one above, on the day.

Closes #265


GigaHierz:
Merge gate met. `setFeeRate(750)` landed on both revealed maps:

| Map | Proxy | Tx | Block |
|---|---|---|---|
| World (0) | `0xA8cFC1B4365518f56954382B6Fab25a5382f5C49` | [`0x5ce4a8…001b`](https://celoscan.io/tx/0x5ce4a84de1dc705701137caad8f4ce1c0eb291a7405f019507a586d186bb001b) | 78274231 |
| Africa (1) | `0x8e70ada33714C3F8f35182b781C63449c5e079b7` | [`0xd2a9e2…40d9`](https://celoscan.io/tx/0xd2a9e2450c2a68f184d14d2ab602774a03e9afe9c61cf801c8ecdcfddb9b40d9) | 78274241 |

Verified from the receipts: status `0x1`, `to` is the proxy above, input selector `0x45596e2e` = `setFeeRate(uint256)` with `0x2ee` = 750, and each emits `FeeRateUpdated(750)`.

Live `feeRate()` at block 78276017 via `cast call <proxy> "feeRate()(uint256)"`:

```
World  0xA8cF…5C49  750
Africa 0x8e70…79b7  750
Asia   0x9b8D…0661  500
maps 3–7            500
```

`https://www.mondeto.app/api/reveals` → `{"ids":[0,1]}`, so every map a player can currently see is at 7.5% and the flat FAQ wording is true today. Asia and the other five maps are still at 500 bps; any of them revealed before its own `setFeeRate` call would make the FAQ wrong for that map. Also: new maps may launch at a higher rate, in which case Terms §6 and both FAQ answers need another pass. Tracking that as a follow-up on #265.

GigaHierz (approved):
APPROVE — reviewed at head `985a344` (= `da3caaf` + merge of main; content diff unchanged).

Verified:
- Diff touches exactly the three flagged strings (Terms §6, FAQ `who-pays-whom`, FAQ `sniped`) plus the two source comments; Terms wording is "up to seven and a half percent (7.5%)".
- `faq.test.ts` pins `7.5% ` and rejects a bare `5% ` (regex excludes `7.5%` and `x.5%`), so a partial revert goes red.
- Runtime code does not hardcode the fee: `lib/resaleFee.ts` reads `feeRate()` per map; analytics and PnL routes go through it. Remaining `500` literals are pure-function test fixtures.
- On-chain: World and Africa at 750 bps, both `setFeeRate` txs decoded and linked in the comment above; production reveals `[0,1]`, so the copy matches every visible map.
- `ci / ci` green on `985a344`; merge-tree against main clean.

Legal sign-off on §6 is being handled by the product owner outside this PR. Follow-up (maps revealed at a different rate) noted on #265.

